PR #39, opened 2021-11-30
First round of small accessibility tweaks.

alexstine:
Small heading adjustments and accessible tab implementation.

May need CSS work to keep same visual look. This is something I'm unable to do myself due to my lack of vision.


partials/welcome-page.php
```diff
@@ -4,11 +4,11 @@
 		<div class="edac-welcome-header">
 			<div class="edac-welcome-header-right">
 				<a href="https://equalizedigital.com/" target="_blank">
-					<img src="<?php echo plugin_dir_url( __DIR__ ); ?>assets/images/edac-logo.png" alt="Equalize Digital">
+					<img src="<?php echo plugin_dir_url( __DIR__ ); ?>assets/images/edac-logo.png" alt="Equalize Digital Logo">
```
amberhinds, line 7: Changing the alt text to "Equalize Digital logo" is not correct because the image is linked to go to the home page of our website, not to the logo.

If we wanted to make it more clear then it would make more sense for the alt to read "Go to Equalize Digital's website"

partials/welcome-page.php
```diff
@@ -46,11 +46,11 @@
 		<div class="edac-welcome-section edac-welcome-section-documentation-support <?php if(get_transient( 'edacp_license_valid' ) == false) echo 'edac-show-pro-callout'; ?>">
 			<?php if(get_transient( 'edacp_license_valid' ) == false){ ?>
 				<div class="edac-welcome-pro-callout">
-					<?php include('pro-callout.php'); ?>
+					<?php include('pro-callout-h4.php'); ?>
```
amberhinds, line 49: I don't think that the filename needs to be changed here does it?

partials/welcome-page.php
```diff
@@ -4,11 +4,11 @@
 		<div class="edac-welcome-header">
 			<div class="edac-welcome-header-right">
 				<a href="https://equalizedigital.com/" target="_blank">
-					<img src="<?php echo plugin_dir_url( __DIR__ ); ?>assets/images/edac-logo.png" alt="Equalize Digital">
+					<img src="<?php echo plugin_dir_url( __DIR__ ); ?>assets/images/edac-logo.png" alt="Equalize Digital Logo">
 				</a>
 			</div>
 			<div class="edac-welcome-header-left">
-				<h2 class="edac-welcome-title">
+				<h1 class="edac-welcome-title">
```
amberhinds, line 11: @SteveJonesDev @alexstine  I expect we might need to have some CSS changes to support these bumped-up heading levels to ensure that the headings are not too big now.

partials/welcome-page.php
alexstine: @amberhinds So ALT text should try to convey what the image is unless it is purely decretive.

https://www.w3.org/TR/WCAG20-TECHS/H37.html

Just saying "Equalize Digital" really doesn't tell me what the image is and I admit, "Equalize Digital Logo" isn't much better.

For the best exposure, link text should probably be added for screen readers, a good image description generated, and image moved below heading 1 main heading.

Thoughts about that?

amberhinds: If it's a linked image I usually make the alt text describe the purpose of the image in the link rather than literally what the image is.  

Like a Facebook icon that links to a Facebook page. It's more useful to hear "Link. Image. Visit our Facebook page" than "Link. Image. Facebook icon." 

I think the same would apply here. "Link. Image. Equalize Digital" makes sense because it goes to the homepage of our website. Whereas "Link. Image. Equalize Digital logo" might make someone think it is a link to an image file of the logo. If we wanted to be more clear then I think we just add more context around the purpose of the links and where it goes.

Arguably, though, we might add aria-hidden="true" and a negative tab index to hide it from screen readers completely since it goes to the same place as the text link that follows it and is in that sense unnecessarily repetitive.

alexstine: I adjusted the ALT text, passing now?

I always go back and forth about this. Honestly, picky Alex would rather never here any mention of an image and just hear "link, Facebook." That's just me though, I think images are for the most part distracting and very rarely adds context for me.

PR #80, opened 2022-12-02
add context stream filter to allow disabling ssl verify

NicktheGeek:
When used in specific environments, the file_get_html() function fails because it cannot verify the SSL certificate. 

This adds a filter that allows the SSL certificate verification to be bypassed.

I did consider including an option to allow this, but because of potential security risk, it seems more prudent to make this a developer-level change.

To use add the filter `add_filter( 'edac_no_verify_ssl', '__return_true' );`. This will include an extra context arg.

This code change accounts for the pro activation authorization functionality but allows local or staging environments where the SSL certificate authentication may fail to enable plugin usage.

Fixes https://github.com/equalizedigital/accessibility-checker/issues/81

### DIFF
--- a/includes/validate.php
+++ b/includes/validate.php
@@ -191,19 +191,49 @@ function edac_get_content($post)
 	// set transient to get html from draft posts
 	set_transient('edac_public_draft',true, 5 * MINUTE_IN_SECONDS);
 
+	/**
+	 * Indicates file_get_html should not verify SSL.
+	 * 
+	 * For site security it is not recommended to use this filter in production.
+	 * 
+	 * @param bool $no_verify_ssl The boolean to check.
+	 */
+	$no_verify_ssl = apply_filters( 'edac_no_verify_ssl', false );
+
 	// http authorization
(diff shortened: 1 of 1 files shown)


includes/validate.php
```diff
@@ -191,19 +191,49 @@ function edac_get_content($post)
...
 			$content['html'] = file_get_html(get_the_permalink($post->ID).'?c='.time(), false, $context);
 		}else{
-			$content['html'] = file_get_html(get_the_permalink($post->ID).'?c='.time());
+			$context = stream_context_create(array(
+				'ssl'  => array(
+					'verify_peer'      => false,
+					'verify_peer_name' => false,
+				),
+			));
+			$content['html'] = file_get_html(get_the_permalink($post->ID).'?c='.time(), false, $context);
```
SteveJonesDev, line 236: I think this shouldn't have a context. This would disable SSL even with the edac_no_verify_ssl filter being set to false.

SteveJonesDev (changes requested):

SteveJonesDev:
Added to version 1.3.9

PR #128, opened 2023-03-24
Update accessibility-checker.php

joedolson:
I know this is super minor. But, just noticed it and thought...what the heck!

Spell 'plugin' correctly. ;)


SteveJonesDev:
Thanks for the spell check @joedolson. 😀

PR #148, opened 2023-05-02
fixed - a video is present duplicated warning issue

vukvukovich:
There were a few additional approaches to this problem that I considered. However, I ultimately chose the most readable solution that maintained the current algorithm's complexity. Implementing other approaches would have required significant refactoring, which I believe was not the main intent of this task.

Please note that there are a couple of minor edge cases not addressed in the current implementation:

1. The first one is having the same video embedded twice with different YouTube URLs (e.g., `youtube.com/foo` and `youtu.be/foo`). While this issue may not arise with Gutenberg, it could potentially occur with other builders. For now, I don't recommend complicating the code with URL normalization, but it's something to keep in mind for future improvements.
2. Another minor edge case is when multiple `<iframe>` elements are found within a single `<figure>` element. Although this is not a recommended practice, it could still happen in certain situations. The current implementation treats each iframe separately, and it does not attempt to consolidate them or handle them differently when they are contained within the same `<figure>`. While this is likely not a significant concern, it's worth considering if you encounter such scenarios or if you decide to further refine the code in the future.

### DIFF
--- a/includes/rules/video_present.php
+++ b/includes/rules/video_present.php
@@ -1,6 +1,6 @@
 <?php
 /**
- * Accessibility Checker pluign file.
+ * Accessibility Checker plugin file.
  *
  * @package Accessibility_Checker
  */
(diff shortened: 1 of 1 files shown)


includes/rules/video_present.php
```diff
@@ -1,6 +1,6 @@
 <?php
 /**
- * Accessibility Checker pluign file.
+ * Accessibility Checker plugin file.
```
SteveJonesDev, line 3: Good catch. 😉

SteveJonesDev (approved):

includes/rules/video_present.php
vukvukovich: Thanks @SteveJonesDev. Do you want me to merge it?

SteveJonesDev: No, I will merge it later. Thanks!

PR #149, opened 2023-05-02
feature - missing subheadings check requirements

vukvukovich:
I have added the minimum word count and professional version requirements.

**Some stuff worth considering:**

- If we might have similar word-limiting occasions this should be pulled out to a helper function. 

- I did not have access to a valid key at the time of testing, so I wasn't able to conduct a thorough test. This pro check should properly work though. Also regarding the check for professional plugin version rules, I suggest that it be implemented separately from the rule-checking functions. One possible solution would include the key with a boolean value for each rule within an individual rule array, in the `edac_register_rules()` function. However, this might make it easy to crack/null the plugin but that is for a different discussion. 

My general suggestion is to deal with the pro check in a different ticket.

### DIFF
--- a/includes/rules/missing_headings.php
+++ b/includes/rules/missing_headings.php
@@ -1,6 +1,6 @@
 <?php
 /**
- * Accessibility Checker pluign file.
+ * Accessibility Checker plugin file.
  *
  * @package Accessibility_Checker
  */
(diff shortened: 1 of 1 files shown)


includes/rules/missing_headings.php
```diff
@@ -13,12 +13,17 @@
  * @return array
  */
 function edac_rule_missing_headings( $content, $post ) {
+	if ( ! EDAC_KEY_VALID ) {
```
SteveJonesDev, line 16: This rule should run on the free plugin. It may have been confusing on the ticket where it said "set this up so that it can be a future setting in Pro". I believe what was meant by that was that the word count of 400 could be set in the Pro plugin settings. But that is outside the scope of this issue. The EDAC_KEY_VALID check can be removed.

includes/rules/missing_headings.php
```diff
@@ -13,12 +13,17 @@
...
+		return;
+	}
 
-	$dom = str_get_html( $post->post_content );
-	if ( empty( $dom ) ) {
-		goto error;
+	$word_count = str_word_count( strip_tags( $post->post_content ) );
+
+	if ( $word_count <= 400 ) {
```
SteveJonesDev, line 22: I think this should only have a less-than sign. The issue called for fewer than 400 words.
"I don't think this should flag if there are fewer than 400 words on the page."

SteveJonesDev (changes requested):

SteveJonesDev (approved):
Looks good! I will merge it later. Thanks!

PR #269, opened 2023-09-06
Fixes #245

### LINKED ISSUE #245: Issue Density array offsets need to be checked if they exist
This happens on a new post because the summary doesn't exist yet.

https://github.com/equalizedigital/accessibility-checker/blob/master/accessibility-checker.php#L1000-L1001

Screenshot 2023-08-29 at 4 08 42 PM

boonedev:
Replicated the issue using:

$issue_density_array = get_post_meta( 5000, '_edac_density_data' );

confirmed that after this fix the warning is no longer shown.


SteveJonesDev (approved):
Tested on my end and it looks to be fixed. Thanks for the PHPCS fixes as well.
Good to merge.

PR #280, opened 2023-09-11
279 edac_remove_elements fails if dom is empty

boonedev:
Helper function edac_remove_elements failed to check if dom was empty before trying to remove elements.


SteveJonesDev (approved):
Good to merge and complete the issue.

PR #290, opened 2023-09-12
Fixes #247

### LINKED ISSUE #247: Can TextStatistics be added via composer
I don't think we can add Simple Dom HTML via composer since we've modified it.

boonedev:
- Adds davechild/textstatistics to composer
- Adds `composer install` to `npm run build`
- Adds loading vendor/autoload.php
- Gracefully fails if davechild/textstatistics does not exist


SteveJonesDev (approved):
Good to merge.

PR #292, opened 2023-09-12
Fixes #253

### LINKED ISSUE #253: Issues_Query Coding Standards
Screenshot 2023-08-29 at 9 57 21 PM

boonedev:
- phpcs updates.


SteveJonesDev (changes requested):
There are some warnings that can be fixed with PHPCBF. I don't think we need to fix warnings related to database cache and direct calls right now.

PR #294, opened 2023-09-12
Fixes #255

### LINKED ISSUE #255: Browser Class
This file:
https://github.com/equalizedigital/accessibility-checker/blob/master/includes/classes/browser.php

Is used in the system info file:
https://github.com/equalizedigital/accessibility-checker/blob/master/includes/system-info.php

It is full of WPCS issues (443 ERRORS AND 111 WARNINGS AFFECTING 276 LINES).

This class wasn't created by us. I don't think it's worth our time to fix all the issues. What are your thoughts on this? Is there a better way to get this information? Do we even need this information? Honestly, I've never referenced browser information when troubleshooting people's support inquiries.

boonedev:
- Adds cbschuld/browser.php to composer (passes phpcs)
- Adds loading from vendor/autoload.php
- Gracefully fails if cbschuld/browser.php does not exist


accessibility-checker.php
```diff
@@ -27,6 +27,11 @@
 // Include plugin dependency.
 require_once ABSPATH . 'wp-admin/includes/plugin.php';
 
+// Load composer packages.
+if ( file_exists( plugin_dir_path( __FILE__ ) . 'vendor/autoload.php' ) ) {
+	include_once plugin_dir_path( __FILE__ ) . 'vendor/autoload.php';
```
SteveJonesDev, line 32: Can we limit the loading of this only to the admin? I'm not sure if there is anything that autoloads needed on the frontend.

SteveJonesDev (changes requested):
User Browser info isn't working now. Even after running the build script.

Master:
Screenshot 2023-09-13 at 1 30 50 PM

This PR:
Screenshot 2023-09-13 at 1 31 13 PM

accessibility-checker.php
boonedev: `npm run build` wasn't updating composer. I pushed an update to dev that adds `composer update` and all looks to be working ok now.

boonedev: It's also now only loading vendor/autoload.php if is_admin.

SteveJonesDev (approved):
Good to merge.

PR #296, opened 2023-09-13
Fixes #260

### LINKED ISSUE #260: Frontend Highlighting: Trigger page scan from frontend
We might need to figure out the best way to achieve this.

Ticket reference: https://wordpress.org/support/topic/front-end-widget-is-unable-to-detect-issues/

boonedev:


accessibility-checker.php
```diff
@@ -782,6 +783,27 @@ function edac_register_rules() {
...
+	
+	// If the admin_bar is visible and we're not editing the post 
+	// assume we are on the frontend and the frontend highlighter is loading. 
+	global $pagenow;
+	if ( is_admin_bar_showing() && 'post.php' != $pagenow ) {
+
+		// Check the page if it hasn't already been checked.
+		global $post;
+		$checked = get_post_meta( $post->ID, '_edac_post_checked', true );
+		if ( false === boolval( $checked ) ) {
```
SteveJonesDev, line 801: We need to check if the current post type is set to be scanned in the plugin settings.

SteveJonesDev (changes requested):

accessibility-checker.php
boonedev: I think edac_validate handles that as a pre-check:
```
function edac_validate( $post_ID, $post, $action ) {
        // check post type.
        $post_types = get_option( 'edac_post_types' );
        if ( is_array( $post_types ) && ! in_array( $post->post_type, $post_types, true ) ) {
                return;
        }
```

SteveJonesDev (approved):
This seems to be working now. It's not updating the summary meta but I'll open a different issue for that.
Screenshot 2023-09-14 at 1 51 36 PM

I also noticed a discrepancy between the ignores listed in the admin and frontend but I think that's a different issue.

This is good to merge.

PR #298, opened 2023-09-13
297 reading level update timing bug

boonedev:


SteveJonesDev (changes requested):
The reading level seems to be updating properly now. Thanks!

When creating a new post before it's saved I'm getting these errors. We'll need to fix those first before releasing this hotfix.

Screenshot 2023-09-14 at 1 18 16 PM

boonedev:
Switching from hotfix to release.

PR #314, opened 2023-09-19
312 welcome page and dashboard widget number format

boonedev:
If approved this PR:

- Adds localized formatting helpers 
- Adds formatted results to the scan_stats summary
- Used the formatted results on the dashboard widget
- Updates date on dashboard to use server localization rather than browser
- Uses the formatted results on the welcome page

### DIFF
--- a/includes/classes/class-helpers.php
+++ b/includes/classes/class-helpers.php
@@ -32,4 +32,47 @@ class Helpers {
 
 		return implode( ',', $items );
 	}
+
+
+	/**
+	 * Localizes the format of a number.
+	 *
+	 * @param [type]  $number
+	 * @param integer $decimals number of decimals.
+	 * @return integer
+	 */
+	public static function format_number( $number, $precision = 0 ) {
+
+		if ( ( ! is_numeric( $number ) ) ) {
+			return $number;
+		}
+		
+		$locale = get_locale();
+
+		$formatter = new \NumberFormatter( $locale, \NumberFormatter::DECIMAL );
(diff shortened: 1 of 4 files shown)


SteveJonesDev (changes requested):
The dates should be pulled from the WordPress settings. Not from the server time.
/wp-admin/options-general.php

The current_time() function is often used in WordPress to retrieve the current time in the format desired and, importantly, it can account for the WordPress site's timezone settings.
WP Docs: https://developer.wordpress.org/reference/functions/current_time/

boonedev:
@SteveJonesDev Isn't it timezone adjusted? Helpers:: format_date accepts a unixtime, then calls WP's date_i18n() that should convert it to the wp_timezone() and format it in the date format setup in the WP option.

SteveJonesDev:
@boonedev, when I change the timezone and run a new full site scan the date doesn't change. Does it change for you?

SteveJonesDev:
I even modified the date output format to output the time to ensure it wasn't changing.

boonedev:
@SteveJonesDev Thanks Steve. [date_i8n](https://developer.wordpress.org/reference/functions/date_i18n/) applies timezone, but does not accept a standard unix timestamp. I've replaced with code that does and now things seem to be  working correctly:

Screenshot 2023-09-20 at 3 56 44 PM

Screenshot 2023-09-20 at 4 00 36 PM

SteveJonesDev (approved):
Dates and timezones are never fun. 😜 Good to merge.

PR #320, opened 2023-09-21
Fixes #216

### LINKED ISSUE #216: Front End Highlighting: using Disable Styles with combined styles breaks app's styling
If the WordPress css is combined as an optimization (ie. Autoptimize plugin's Aggregate CSS-files option), Disable Styles will break app's styling:

Screenshot 2023-07-28 at 9 58 00 AM

We likely need to inline app's styles. See: https://blog.logrocket.com/the-best-react-inline-style-libraries-comparing-radium-aphrodite-emotion-849ef148c473/

boonedev:
I kept trying to make this hard...

(Autoptimize plugin)
Screenshot 2023-09-21 at 4 31 31 PM

**Before w/styles disabled**
Screenshot 2023-09-21 at 4 32 10 PM

**After w/styles disabled**
Screenshot 2023-09-21 at 4 32 30 PM

### DIFF
--- a/src/app/index.js
+++ b/src/app/index.js
@@ -699,6 +699,25 @@ class AccessibilityCheckerHighlight {
 	 * This function disables all styles on the page.
 	 */
 	disableStyles() {
+
+		/*
+		If the site compiles css into a combined file, our method for disabling styles will cause out app's css to break.
+		This checks if the app's css is loading into #edac-app-css as expected. 
+		If not, then we assume the css has been combined, so we manually add it to the document.
+		*/
+		if( ! document.querySelector('#edac-app-css') ){
+			debug('css is combined, so adding app.css to page.');
+
+			var link = document.createElement('link');
+			link.rel = 'stylesheet';
+			link.id = 'edac-app-css';
+			link.type = 'text/css';
+			link.href = edac_script_vars.appCssUrl,
+			link.media = 'all';
+			document.head.appendChild(link);
+		}
+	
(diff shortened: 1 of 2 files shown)


src/app/index.js
```diff
@@ -699,6 +699,25 @@ class AccessibilityCheckerHighlight {
...
 	 */
 	disableStyles() {
+
+		/*
+		If the site compiles css into a combined file, our method for disabling styles will cause out app's css to break.
+		This checks if the app's css is loading into #edac-app-css as expected. 
+		If not, then we assume the css has been combined, so we manually add it to the document.
+		*/
+		if( ! document.querySelector('#edac-app-css') ){
+			debug('css is combined, so adding app.css to page.');
```
SteveJonesDev, line 709: Does the build script remove debug calls?

SteveJonesDev (changes requested):
A nice simple solution. Works well, I just have a question.

src/app/index.js
boonedev: It doesn't. The  debug and info functions just check for  INFO_ENABLED & DEBUG_ENABLED flag variables before they output anything to console. I think uglify removes console and comments though if you want to go that route?

SteveJonesDev: It's not necessary right now. Maybe something to think about in the future. There might be a performance benefit in doing so.

SteveJonesDev (approved):

PR #322, opened 2023-09-22
321 passed tests should not be able to be negative

boonedev:
If approved this PR will:

- Update passed_percentage calculation


SteveJonesDev (approved):
Looks good. Please merge.

PR #327, opened 2023-09-24
Hotfix/304 empty linkphp rule logic false positive

boonedev:
If approved this hotfix PR will:
- Update edac_get_body_density_data to receive html rather than $dom object
- Update logic in empty_link.php 
- minor phpcs spacing changes


SteveJonesDev (approved):
Good to merge.

Issue #324, opened 2023-09-22
Add Email Opt-in


SteveJonesDev:
Details: https://3.basecamp.com/3579237/buckets/29333825/messages/6574975590#__recording_6575357951

boonedev:
@SteveJonesDev should this be in its own panel on should it appear inside the Learn Accessiblity panel? Also, should the confirmation message be inline or in a modal?

amberhinds:
<img width="602" alt="Screenshot 2023-09-25 at 10 26 25 AM" src="https://github.com/equalizedigital/accessibility-checker/assets/13153456/5d16e547-b858-4155-ab65-6c774a5e938c">
@boonedev

boonedev:
The zoho confirmation redirects to a new page by default. @amberhinds could you set the form's response to inline? ie:

<img width="755" alt="Screenshot 2023-09-25 at 3 45 30 PM" src="https://github.com/equalizedigital/accessibility-checker/assets/695220/7a6f941f-88ed-4271-9b05-59c8cbeaea20">

christophermhinds:
@boonedev @SteveJonesDev form settings are now inline and should display an inline confirmation for Focus State sign-up.

PR #329, opened 2023-09-26
Adds #324

boonedev:
If approved this PR:
- Adds email opt-in panel 
- Fixes misc phpcs issues

### DIFF
--- a/includes/classes/class-welcome-page.php
+++ b/includes/classes/class-welcome-page.php
@@ -253,9 +253,6 @@ class Welcome_Page {
 							<div class="SIGNUP_FLD edac-mt-3" id="edac-opt-in-email">
 								<label style="font-weight: 600;" for="EMBED_FORM_EMAIL_LABEL">Email Address (Required)</label>
 								<input style="max-width: 100%;" type="email" changeitem="SIGNUP_FORM_FIELD" name="CONTACT_EMAIL" id="EMBED_FORM_EMAIL_LABEL" aria-describedby="email-info-region" value="' . esc_attr( $email ) . '">
-								<div class="edac-mt-3" id="email-info-region" class="info-message">
-									' . esc_html( __( 'Note: The Email Address field has been auto-populated with your WordPress email.', 'accessibility-checker' ) ) . '
-								</div>
 							</div>
 							<div class="edac-mt-0" style="display: none;" id="errorMsgDiv">' . __( 'Please enter a valid email address.', 'accessibility-checker' ) . '</div>
 						
@@ -308,15 +305,6 @@ class Welcome_Page {
 			
 			<script>
 
-				const email_element = document.querySelector("#EMBED_FORM_EMAIL_LABEL");
-				const email_info_region = document.querySelector("#email-info-region");
-				const submit_element = document.querySelector("#zcWebOptin");
-
-				email_element.addEventListener("input", (event) => {
-					 email_info_region.textContent = "";
-				});
-				
-			
 				function initOnComplete(){
 					const optIn = document.querySelector("#edac-opt-in-email");
 
--- a/includes/classes/class-welcome-page.php
+++ b/includes/classes/class-welcome-page.php
@@ -285,7 +285,7 @@ class Welcome_Page {
 							</form>
 
 							<!-- Confirmation Message -->
-							<div id="confirmationMessage" tabindex="-1" role="alert" style="outline: none;">
+							<div id="confirmationMessage" tabindex="-1" style="outline: none;">
 							<!-- This will be populated with the confirmation message on form submit -->
 							</div>
 						</div>
--- a/includes/classes/class-welcome-page.php
+++ b/includes/classes/class-welcome-page.php
@@ -218,30 +218,45 @@ class Welcome_Page {
 	 */
 	public static function render_email_opt_in( $zx, $zcld, $zctd, $zc_form_ix ) {
 		
+		$current_user = wp_get_current_user();
+		$email        = $current_user->user_email;
+
+		if ( empty( $email ) ) {
+			$email = '';
+		}
+
 		$form_id = 'sf' . $zc_form_ix;
 
 		$html = '
 			<script type="text/javascript" src="https://zmp-glf.maillist-manage.com/js/optin.min.js" onload="setupSF(\'' . esc_attr( $form_id ) . '\',\'ZCFORMVIEW\',false,\'light\',false,\'0\')"></script>
 			<script type="text/javascript">
+				var _edac_email_opt_in_email = "";
 				function runOnFormSubmit_' . esc_html( $form_id ) . '(th){
-					/*Before submit, if you want to trigger your event, "include your code here"*/
+					_edac_email_opt_in_email = document.querySelector("#EMBED_FORM_EMAIL_LABEL").value;
 				};
 			</script>
 
-			<div class="edac-panel edac-mt-1" id="' . esc_attr( $form_id ) . '" data-type="signupform" style="opacity: 1;">
-				<div id="customForm">
+			<div class="edac-panel edac-mt-1 edac-pb-3">
+			<div id="' . esc_attr( $form_id ) . '" data-type="signupform" style="opacity: 1;">
+
+			<div id="customForm">
 					<div class="quick_form_8_css" name="SIGNUP_BODY">
-						<div class="opt-in-hide">
+						<div>
 							<h2 id="SIGNUP_HEADING">' . __( 'Get Notified of Upcoming Events', 'accessibility-checker' ) . '</h2>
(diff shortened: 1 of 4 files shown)


SteveJonesDev:
@boonedev, is the form supposed to disappear after being filled out? Refreshing the page doesn't bring it back either.

SteveJonesDev:
Looks like it's per user.

boonedev:
Sounds like a bug. It should:
1) show the form
2) if user email submit is good, the form should thank them, then hide the form and not display again.

Which browser?

boonedev:
Also, what do you think about pre-populating the form with their email as a convenience?

SteveJonesDev:
Using the Brave browser.

1. The form is shown
2. I enter an email address
3. It shows a thank you message
4. Then after about a second, the form is dynamically removed from the dom

boonedev:
So thank-you/hide happens before the submit?

SteveJonesDev:
Sorry, no I missed a step. I did click the submit button. I don't think the form should be dynamically removed from the page. I'm not sure if you're doing that or Zoho.

includes/classes/class-welcome-page.php
```diff
@@ -205,4 +205,144 @@ public static function render_summary() {
...
+							<span style="display: none" id="dt_CONTACT_EMAIL">1,true,6,Contact Email,2</span>
+							</form>
+						</div>
+						<div class="opt-in-show" style="display: none;">
+							<h2>' . __( 'Get Notified of Upcoming Events', 'accessibility-checker' ) . '</h2>
+							<div>' . __( 'Thank-you for signing up!', 'accessibility-checker' ) . '</div>
+						</div>
+					</div>
+				</div>
+				<img src="https://zmp-glf.maillist-manage.com/images/spacer.gif" id="refImage" onload="referenceSetter(this)" style="display:none;">
```
SteveJonesDev, line 278: Accessibility: A layout spacer image (which should have null/empty alternative text) does not have an alt attribute.

boonedev:
You've got it. Ok to not show the panel on page refresh if they have already signed-up, or should it always show the thank-you message?

includes/classes/class-welcome-page.php
```diff
@@ -205,4 +205,144 @@ public static function render_summary() {
...
+						<div class="opt-in-hide">
+							<h2 id="SIGNUP_HEADING">' . __( 'Get Notified of Upcoming Events', 'accessibility-checker' ) . '</h2>
+							<div>' . __( 'Join our email list and get event reminders and recordings in your inbox.', 'accessibility-checker' ) . '</div>
+							<div>
+								<div id="Zc_SignupSuccess" style="display: none;">' . __( 'Thank-you for signing up!', 'accessibility-checker' ) . '</div>
+							</div>
+							<form method="POST" id="zcampaignOptinForm" style="margin: 0px; width: 100%" action="https://zmp-glf.maillist-manage.com/weboptin.zc" target="_zcSignup">
+							
+							<div class="SIGNUP_FLD edac-mt-1" id="edac-opt-in-email">
+								<label style="font-weight: 600;" for="EMBED_FORM_EMAIL_LABEL">Email Address *</label>
```
SteveJonesDev, line 243: Accessibility: Change asterisk to '(Required)'

SteveJonesDev (changes requested):
There are two hidden inputs at the bottom of the form that don't have form labels. I can't find them in the PHP so I'm assuming they are being added via JavaScript. I'm not sure if we can fix that or not.
Screenshot 2023-09-27 at 11 20 47 AM

Can we auto-populate the email input with the current logged-in user's email address?
If we do auto-populate we might want to inform screen readers.
Example:
```
<form>
  <label for="username">Username</label>
  <input type="text" id="username" aria-describedby="info-region" />
  <div id="info-region" aria-live="polite" class="info-message"></div>
</form>
```

```
document.addEventListener("DOMContentLoaded", function () {
  const usernameInput = document.getElementById('username');
  const infoRegion = document.getElementById('info-region');
  
  usernameInput.value = 'AutoPopulatedUser';
  infoRegion.textContent = 'Note: Username field has been auto-populated with "' + usernameInput.value + '".';
});
```

boonedev:
The extra inputs are added by zoho's js onload. I could try to listen for them to exist then replace them with hidden fields if you think it's worth it?

Updates applied:
- no longer removes fields on submit, instead makes the email field readonly and disabled, and disables the submit button
- auto-populates WP email
- adds "Note: The Email Address field has been auto-populated with your WordPress email." notice & removes it when the email field is modified
- adds missing alt for spacer image

SteveJonesDev (changes requested):
I think the a11y of this is worse. I think removing the form after submission is fine and showing the confirmation message. But we need to ensure the screen reader focuses on that message and reads it out on submit. Let's remove the aria-live message for now.

Here is an example of the confirmation message.
```
<!-- Form HTML -->
<form id="myForm">
  <!-- Form fields go here -->
  <input type="submit" value="Submit" />
</form>

<!-- Confirmation Message -->
<div id="confirmationMessage" tabindex="-1" role="alert" style="outline: none;">
  <!-- This will be populated with the confirmation message on form submit -->
</div>

<script>
// JavaScript to handle form submit
document.getElementById('myForm').addEventListener('submit', function(e){
  e.preventDefault(); // Prevent actual form submit
  
  // Populate the confirmation message
  document.getElementById('confirmationMessage').textContent = 'Your form has been submitted successfully.';
  
  // Set focus to the confirmation message
  document.getElementById('confirmationMessage').focus();
});
</script>
```

boonedev:
So leave as-is and only remove `aria-describedby` and ` aria-live` attributes below:
```
<input name="CONTACT_EMAIL" id="EMBED_FORM_EMAIL_LABEL" aria-describedby="email-info-region">
<div class="edac-mt-3" id="email-info-region" aria-live="polite" class="info-message"></div>
```
Or re-add removing the form on submit and show a confirm message?

SteveJonesDev:
This: Or re-add removing the form on submit and show a confirm message?

But only the form, not the whole subscribe container like it did before. Test it with Voice Over and ensure it reads out the confirmation message on submit.

If you need an example install Gravity Forms, create a test form, and set it to submit via Ajax.

boonedev:
Updated.
- Adds confirmation message with role=alert 
- On submit, removes the form
 - On submit, populates confirmation message text and gives it focus
 - Tested w/Voice Over -on submit confirmation message is read-out

includes/classes/class-welcome-page.php
```diff
@@ -205,4 +205,184 @@ public static function render_summary() {
...
+								<div id="Zc_SignupSuccess" style="display: none;"></div>
+							</div>
+							
+							<form method="POST" id="zcampaignOptinForm" style="margin: 0px; width: 100%" action="https://zmp-glf.maillist-manage.com/weboptin.zc" target="_zcSignup">
+							
+							<div class="SIGNUP_FLD edac-mt-3" id="edac-opt-in-email">
+								<label style="font-weight: 600;" for="EMBED_FORM_EMAIL_LABEL">Email Address (Required)</label>
+								<input style="max-width: 100%;" type="email" changeitem="SIGNUP_FORM_FIELD" name="CONTACT_EMAIL" id="EMBED_FORM_EMAIL_LABEL" aria-describedby="email-info-region" value="' . esc_attr( $email ) . '">
+								<div class="edac-mt-3" id="email-info-region" class="info-message">
+									' . esc_html( __( 'Note: The Email Address field has been auto-populated with your WordPress email.', 'accessibility-checker' ) ) . '
```
SteveJonesDev, line 257: Delete this message.

includes/classes/class-welcome-page.php
```diff
@@ -205,4 +205,184 @@ public static function render_summary() {
...
+							<input type="hidden" id="viewFrom" value="URL_ACTION">
+							<input type="hidden" name="zx" id="cmpZuid" value="' . esc_attr( $zx ) . '">
+							<input type="hidden" id="zcld" name="zcld" value="' . esc_attr( $zcld ) . '">
+							<input type="hidden" id="zctd" name="zctd" value="' . esc_attr( $zctd ) . '">
+							<input type="hidden" id="zc_formIx" name="zc_formIx" value="' . esc_attr( $zc_form_ix ) . '">
+							<span style="display: none" id="dt_CONTACT_EMAIL">1,true,6,Contact Email,2</span>
+							</form>
+
+							<!-- Confirmation Message -->
+							<div id="confirmationMessage" tabindex="-1" role="alert" style="outline: none;">
```
SteveJonesDev, line 288: The confirmation message is being read out twice by the screen reader. Remove role="alert" and test that it's only being read out once.

SteveJonesDev (changes requested):

boonedev:
Removed role=alert and confirmed message is announced only once.

SteveJonesDev (approved):
Please merge.

PR #364, opened 2023-11-22
Hotfix to handle loopback environments

boonedev:
- Adds check if server file_get_contents calls to the site domain are to a loopback ip.
- Adds logic to not validate ssl on loopback calls.

### DIFF
--- a/includes/validate.php
+++ b/includes/validate.php
@@ -275,29 +275,13 @@ function edac_get_content( $post ) {
 	$is_local_loopback = get_option( 'edac_local_loopback', null );
 	
 	if ( null === $is_local_loopback ) {
-
+	
 		$parsed_url = wp_parse_url( home_url() );
 
 		if ( isset( $parsed_url['host'] ) ) {
-			
-			// Get the ip of the domain.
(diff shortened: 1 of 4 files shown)


includes/validate.php
```diff
@@ -263,14 +263,52 @@ function edac_get_content( $post ) {
...
+	
+	if ( null === $is_local_loopback ) {
+
+		$parsed_url = wp_parse_url( home_url() );
+
+		if ( isset( $parsed_url['host'] ) ) {
+			
+			// Get the ip of the domain.
+			$site_domain = $parsed_url['host'];
+			$host_ip     = gethostbyname( $site_domain );
```
SteveJonesDev, line 285: Do we need to be concerned with IPv6 as `gethostbyname()` doesn't support IPv6 addresses? `dns_get_record()` might be an alternative function if it's something we should be concerned with.

SteveJonesDev (changes requested):

SteveJonesDev (approved):

SteveJonesDev (approved):

PR #365, opened 2023-11-27
Fix: Correct callback func for rest endpoint

jawadhere:
Rest API route : /scans-stats-by-post-type/ is not working as expected.

This same route with same ( GET ) endpoint is declared twice with different callbacks
As a result the first callback function is getting executed on calls which is already used 
in /scans-stats/ route.

If this PR approves it will: 

 - Add the correct callback function 'get_scans_stats_by_post_type' on '/scans-stats-by-post-type/(?P<slug>[a-zA-Z0-9_-]+)'
 - Remove the duplicate route registration


boonedev:
Rolled into https://github.com/equalizedigital/accessibility-checker/pull/443 so closing.

PR #372, opened 2023-12-01
Run PHPCS in PRs, for modified files

aristath:
This PR adds a workflow in PRs to do the following:
* Checks if there are any PHP files changed in the PR
* If any PHP files were changed, then run `phpcs` to those files.

This way, Coding Standards rules can be applied incrementally. If someone makes a change to a PHP file, then the whole file will be checked and any errors will be reported and required to be fixed before the PR gets merged.

The alternative would be to create a massive PR fixing all CS issues at once, but that would be extremely difficult to properly review. 

Having an action consistently run in PRs will help improve and maintain the code quality of this project 👍

### DIFF
--- a/.github/workflows/phpcs.yml
+++ b/.github/workflows/phpcs.yml
@@ -1,6 +1,8 @@
 name: PHPCS check
 on:
-  push
+  push:
+    paths:
+      - '**.php'
 
 env:
   GH_TOKEN: ${{ github.token }}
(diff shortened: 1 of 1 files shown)


SteveJonesDev:
@aristath, We've already added the PHPCS workflow, as well as PHP Lint and Security workflows from https://github.com/jdevalk/comment-hacks/tree/trunk/.github/workflows. See the `develop` branch which will be merged into `master' shortly.

PR #383, opened 2023-12-06
Coding Standards & code-quality tweaks

aristath:
## What this PR does:
 
* The `%i` placeholder can be used for table-names, and allows us to properly prepare SQL queries.
* Strict comparisons (`===`/`!==` instead of `==`/`!=`) should be used where possible.
* Single-use variables should be avoided. Doing `$sql = 'foo'; $wpdb->query( $sql )` is bad practice and we should prefer `$wpdb->query( 'foo' );` - especially if the `$sql` variable is only used once. There's no reason to fill the PHP memory with things like that, and it makes the codes easier to read and understand.
* There were a few instances where the wrong textdomain was used in strings, so they were not translatable. I added an extra rule in `phpcs.xml` to catch these in the future, and fixed the instances where there was an issue. The textdomains that were used where `accessibility-checker` (correct), `accessibility_checker` (changed to `accessibility-checker`) and `edac` (changed to `accessibility-checker`).

## Process:

* I went through all the `phpcs:ignore` statements in the code and reviewed them. If it was possible to remove the rule ignores, and fix the code instead, this was done (for example by using the `%i` placeholder to properly prepare SQL statements).
* I added the default WP phpcs rules to the `phpcs.xml` file.
* I ran `vendor/bin/phpcs` and fixed any issues reported by that
* I commented out (locally - not committed) the `<rule ref="WordPress-VIP-Go"/>` line in `phpcs.xml`. The VIP rules disable some of the WP default rules for code-quality.
* With the VIP rules disabled, I ran `phpcbf` to automatically fix some of the simple stuff like spaces at the ends of lines, indentation etc. After that, I ran `phpcs` again and fixed the remaining issues.

### DIFF
--- a/includes/classes/class-welcome-page.php
+++ b/includes/classes/class-welcome-page.php
@@ -24,7 +24,6 @@ class Welcome_Page {
 	 */
 	public static function render_summary() {
 
-		$html        = '';
 		$scans_stats = new Scans_Stats();
 		$summary     = $scans_stats->summary();
 		?>
--- a/includes/classes/class-welcome-page.php
+++ b/includes/classes/class-welcome-page.php
@@ -1,8 +1,10 @@
 <?php
 /**
  * Class file for Welcome Page
- * 
+ *
  * @package Accessibility_Checker
+ *
+ * phpcs:disable WordPress.WP.EnqueuedResources.NonEnqueuedScript
  */
 
 namespace EDAC;
@@ -13,7 +15,7 @@ use EDAC\Scans_Stats;
  * Class that handles welcome page
  */
 class Welcome_Page {
-	
+
 
 	/**
 	 * Renders page summary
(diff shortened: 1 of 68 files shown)


includes/classes/class-welcome-page.php
aristath, line 1: Changes in this file might seem big, however it was a relatively simple change:
Instead of doing this:
```php
$html = '<div>' . __( 'foo', 'bar' ) . '</div>';
echo $html;
```
The syntax was switched to this:
```php
?>
<div><?php esc_html_e( 'foo', 'bar' ); ?></div>
<?php
```
This has the following benefits:
* We can properly - and consistently, escape strings where needed
* Proper nesting of elements in-code - makes debugging easier/faster
* IDEs can properly highlight the syntax - making it easier for developers to see what happens where

Additionally, I found a couple of strings that were not translatable so this was also fixed (for example, `Your site has a large number of issues. For performance reasons, not all issues have been included in this summary` was not translatable)

includes/rules/possible_heading.php
```diff
@@ -111,14 +111,14 @@ function edac_css_font_size_weight_check( $content ) {
...
-							if ( in_array( $rules['font-weight'], array( 'bold', 'bolder', '700', '800', '900' ) ) ) {
+							if ( in_array( $rules['font-weight'], array( 'bold', 'bolder', '700', '800', '900' ), true ) ) {
 								$has_bold_or_italic = true;
 							}
 						}
 						if ( array_key_exists( 'font-style', $rules ) ) {
 							// replace CSS variables.
 							$rules['font-style'] = edac_replace_css_variables( $rules['font-style'], $css_array );
-							if ( in_array( $rules['font-style'], array( 'italic', 'oblique' ) ) ) {
+							if ( 'italic' === $rules['font-style'] || 'oblique' === $rules['font-style'] ) {
```
aristath, line 121: `in_array()` is a relatively slow function - like all the array operations in PHP, and should be avoided if possible. This was a simple change since we just want to check if the font-style is one of 2 values.

includes/validate.php
```diff
@@ -184,65 +184,47 @@ function edac_validate( $post_ID, $post, $action ) {
...
 	foreach ( $rules as $rule ) {
-		if ( array_key_exists( 'ruleset', $rule ) && 'js' === $rule['ruleset'] ) {
-			$js_rule_ids[] = $rule['slug'];
-		} else {
-			$php_rule_ids[] = $rule['slug'];
+		if ( 'js' === $ruleset && isset( $rule['ruleset'] ) && 'js' === $rule['ruleset'] ) {
+			$rule_slugs[] = $rule['slug'];
+		} elseif ( 'js' !== $ruleset ) {
+			$rule_slugs[] = $rule['slug'];
 		}
```
aristath, line 194: Instead of compiling 2 different arrays, one for JS and one for PHP, this changes the code to only compile a single array with the rules we need. We already know if we need PHP or JS rules since it's passed to the function as `$ruleset`, so it's a lot cleaner to have a single array and then use that in the query. This also helps simplify the code below for queries.

includes/validate.php
```diff
@@ -184,65 +184,47 @@ function edac_validate( $post_ID, $post, $action ) {
...
-	// Build a sql sanitized list from an array
-	// See: https://stackoverflow.com/questions/10634058/wordpress-prepared-statement-with-in-condition .
-	$php_rule_ids = array_map(
-		function ( $v ) {
-			return "'" . esc_sql( $v ) . "'";
-		},
-		$php_rule_ids
-	);
-	$php_rule_ids = implode( ',', $php_rule_ids );
-
```
aristath, line 217: This is no longer needed. Sanitization etc now happens in the query below - which is properly prepared

includes/validate.php
```diff
@@ -184,65 +184,47 @@ function edac_validate( $post_ID, $post, $action ) {
...
+				sprintf(
+					"UPDATE {$wpdb->prefix}accessibility_checker SET recordcheck = %%d WHERE siteid = %%d and postid = %%d and type = %%s AND rule IN (%s)",
+					implode( ',', array_fill( 0, count( $rule_slugs ), '%s' ) )
+				),
+				array_merge(
+					array( 0, get_current_blog_id(), $post_ID, $type ),
+					$rule_slugs
+				)
+			)
+		);
```
aristath, line 212: Breaking down what this does:
* All placeholders are escaped (for example `%%d`), except the last one (`%s`). 
* The `implode( ',', array_fill(...) )` line will take the `$rule_slugs` var, and if it has 3 items it will return `%s,%s,%s`, if it has 4 items it will return `%s,%s,%s,%s` etc.
After `sprintf`, this line:
```php
"UPDATE {$wpdb->prefix}accessibility_checker SET recordcheck = %%d WHERE siteid = %%d and postid = %%d and type = %%s AND rule IN (%s)",
```
gets converted to this (assuming in this example that `$rule_slugs` contains 3 items):
```php
"UPDATE {$wpdb->prefix}accessibility_checker SET recordcheck = %d WHERE siteid = %d and postid = %d and type = %s AND rule IN (%s,%s,%s)",
```
* The `$wpdb->prepare()` method then takes over, and properly quotes `%s` where needed.

The same method is now used in WP-Core to get around issues where it was previously hard to properly prepare queries.

partials/settings-page.php
```diff
@@ -30,19 +30,19 @@
...
 			if ( $a['order'] < $b['order'] ) {
 				return -1;
-			} elseif ( $a['order'] == $b['order'] ) {
+			}
+			if ( $a['order'] === $b['order'] ) {
 				return 0;
-			} else {
-				return 1;
 			}
+			return 1;
```
aristath, line 37: The `else` statement here was simply not necessary because the previous line returns early if the condition is true.

includes/insert.php
```diff
@@ -8,13 +8,13 @@
...
- * @param string $ruletype the rule type.
- * @param string $object the object.
+ * @param object $post     The post object.
+ * @param string $rule     The rule.
+ * @param string $ruletype The rule type.
+ * @param string $rule_obj The object.
  * @return void
  */
-function edac_insert_rule_data( $post, $rule, $ruletype, $object ) { // phpcs:ignore Universal.NamingConventions.NoReservedKeywordParameterNames.objectFound -- Object is a valid parameter name in this context.
+function edac_insert_rule_data( $post, $rule, $ruletype, $rule_obj ) {
```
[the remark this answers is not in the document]
SteveJonesDev: Noted and will watch for these moving forward.

SteveJonesDev (changes requested):
@aristath, there is a merge conflict between your changes to the phpcs.xml file and @jdevalk changes. If you don't mind, can you fix the conflicts so we're sure it's correct?

Also, given your feedback on the WordPress-VIP-Go standard should we not include that in the phpcs.xml?

aristath:
Thank you @SteveJonesDev!

* Merged the latest changes from `develop` to this branch so there are no conflicts now 👍 
* Added back the rules that the VIP-Go ruleset was removing
* Resolved some `unused variable` warnings that were thrown after adding back the WP-Core rules

> given your feedback on the WordPress-VIP-Go standard should we not include that in the phpcs.xml?

It's OK to include it... It adds some VIP-specific rules. Now that I added back some of the WP-Core rules that were being removed, there won't be an issue with it 👍

SteveJonesDev (approved):

PR #401, opened 2023-12-08
Add/phpunit

aristath:


SteveJonesDev:
@aristath, To get the workflows to work down to PHP 7.4 I had to downgrade `doctrine/instantiator` to `^1.3.1`. Let me know if that is okay or what your thoughts are on that.

Thanks for setting this up and creating examples! It's something I've been wanting to do for a long time. 😀

PR #412, opened 2023-12-11
Remove calls to add_option and replaces it with update_option where appropriate

aristath:
Calling `add_option` is in most cases unnecessary. In this PR we're just cleaning up these calls.

### DIFF
--- a/accessibility-checker.php
+++ b/accessibility-checker.php
@@ -253,11 +253,7 @@ function edac_update_database() {
 	// Update database version option.
 	$option_name = 'edac_db_version';
 	$new_value   = EDAC_DB_VERSION;
-	if ( get_option( $option_name ) !== false ) {
-		update_option( $option_name, $new_value );
-	} else {
-		add_option( $option_name, $new_value );
-	}
+	update_option( $option_name, $new_value );
 }
 
 /**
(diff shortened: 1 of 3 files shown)


accessibility-checker.php
```diff
@@ -253,11 +253,7 @@ function edac_update_database() {
 	// Update database version option.
 	$option_name = 'edac_db_version';
 	$new_value   = EDAC_DB_VERSION;
-	if ( get_option( $option_name ) !== false ) {
-		update_option( $option_name, $new_value );
-	} else {
-		add_option( $option_name, $new_value );
-	}
+	update_option( $option_name, $new_value );
```
aristath, line 256: There's no reason to have this condition here. `update_option` takes care of both scenarios for us.

accessibility-checker.php
```diff
@@ -1069,19 +1065,12 @@ function edac_anww_update_post_meta() {
...
-
+	if ( get_option( $option_name ) === false && EDAC_ANWW_ACTIVE ) {
+		update_option( $option_name, true );
+	} elseif ( get_option( $option_name ) === true && ! EDAC_ANWW_ACTIVE ) {
 		delete_option( $option_name );
-
-		edac_update_post_meta( 'link_blank' );
-
 	}
+	edac_update_post_meta( 'link_blank' );
```
aristath, line 1073: * Since `edac_update_post_meta` runs regardless of the outcome of the condition, it can be moved outside it
* Replaced `add_option` with `update_option`

includes/validate.php
```diff
@@ -112,11 +112,10 @@ function edac_validate( $post_ID, $post, $action ) {
 	do_action( 'edac_after_get_content', $post_ID, $content, $action );
 
 	if ( ! $content['html'] ) {
-		add_option( 'edac_password_protected', true );
+		update_option( 'edac_password_protected', true );
 		return;
-	} else {
-		delete_option( 'edac_password_protected' );
 	}
+	delete_option( 'edac_password_protected' );
```
aristath, line 118: * Replaced `add_option` with `update_option`
* The `else` is not necessary since the previous condition bails early

SteveJonesDev (approved):

PR #413, opened 2023-12-11
Review usage of `} else {`

aristath:
In most cases, using `else` is not required and the code can be cleaned-up by removing these.
In this PR I went through some of these and simplified the code where required.

### DIFF
--- a/includes/classes/class-helpers.php
+++ b/includes/classes/class-helpers.php
@@ -55,10 +55,8 @@ class Helpers {
 			$formatter->setAttribute( \NumberFormatter::GROUPING_USED, 1 ); // Include thousands separator.
 
 			return $formatter->format( $number );
-
-		} else {
-			return number_format( $number );
 		}
+		return number_format( $number );
 	}
 
 	/**
@@ -86,10 +84,8 @@ class Helpers {
 			$formatter->setAttribute( \NumberFormatter::MAX_FRACTION_DIGITS, $precision ); // decimals to include.
 
 			return $formatter->format( $number );
-
-		} else {
-			return sprintf( '%.2f%%', $number * 100 );
 		}
+		return sprintf( '%.2f%%', $number * 100 );
 	}
 
 	/**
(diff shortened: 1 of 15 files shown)


includes/classes/class-helpers.php
```diff
@@ -147,14 +137,7 @@ public static function get_option_as_array( $option_name ) {
 
 		if ( is_array( $option ) && ! empty( $option ) ) {
 			return $option;
-		} else {
-			return array();
-		}
-
-		if ( is_string( $option ) ) {
-			return array( $option );
```
aristath, line 155: In this scenario, the `is_string` condition was never reachable - something that became apparent once the previous condition was simplified.

includes/enqueue-scripts.php
```diff
@@ -82,19 +82,15 @@ function edac_enqueue_scripts( $mode = '' ) {
...
 
 	if ( '' === $mode ) {
-		if ( ( current_user_can( 'edit_post', $post_id ) && ! is_customize_preview() )
-		) {
-			$mode = 'ui';
-		} else {
+		if ( ! current_user_can( 'edit_post', $post_id ) || is_customize_preview() ) {
 			return;
 		}
+		$mode = 'ui';
```
aristath, line 88: In this case, reversing the logic of the condition allows us to bail early when needed.

includes/enqueue-scripts.php
```diff
@@ -82,19 +82,15 @@ function edac_enqueue_scripts( $mode = '' ) {
...
 	}
 
 	if ( ( 'full-scan' === $mode && $pro )
-			||
-			( 'ui' === $mode || 'editor-scan' === $mode
-			&&
-			$post_id && current_user_can( 'edit_post', $post_id ) )
+		|| 'ui' === $mode 
+		|| ( 'editor-scan' === $mode && $post_id && current_user_can( 'edit_post', $post_id ) )
```
aristath, line 93: Refactored this condition a bit, because with the prioritization of `||`/`&&` it was not apparent at first glance what it does.

includes/helper-functions.php
```diff
@@ -215,13 +213,10 @@ function edac_is_gutenberg_active() {
...
-		$diff = strtotime( $activation_date ) - strtotime( gmdate( 'Y-m-d H:i:s' ) );
-
-		$days_active = abs( round( $diff / 86400 ) );
-	} else {
-		$days_active = null;
+		$diff        = strtotime( $activation_date ) - strtotime( gmdate( 'Y-m-d H:i:s' ) );
+		return abs( round( $diff / 86400 ) );
 	}
-	return $days_active;
+	return 0;
```
aristath, line 219: This code was sometimes returning `null` instead of an integer, so that was fixed

includes/rules/color_contrast_failure.php
```diff
@@ -346,13 +347,9 @@ function edac_test_color_diff( $color1, $color2 ) {
...
-	if ( $l1 > $l2 ) {
-		$dif = ( ( $l1 + 0.05 ) / ( $l2 + 0.05 ) ) + 0.05;
-	} else {
-		$dif = ( ( $l2 + 0.05 ) / ( $l1 + 0.05 ) ) + 0.05;
-	}
-
-	return $dif;
+	return ( $l1 > $l2 )
+		? ( ( $l1 + 0.05 ) / ( $l2 + 0.05 ) ) + 0.05
+		: ( ( $l2 + 0.05 ) / ( $l1 + 0.05 ) ) + 0.05;
```
aristath, line 352: Technically this is the same as using `else`, but it makes the code easier to read and it's easier to see at a glance what the difference between the 2 actually is 👍

SteveJonesDev (approved):

PR #414, opened 2023-12-11
Remove unnecessary `has_filter()` calls

aristath:
There's no need to check if `has_filter` before applying filters.
This PR removes the cases where `has_filter()` was not required.

### DIFF
--- a/accessibility-checker.php
+++ b/accessibility-checker.php
@@ -1237,10 +1237,7 @@ function edac_details_ajax() {
 	$rules = array_merge( $error_rules, $warning_rules, $passed_rules );
 
 	if ( $rules ) {
-		$ignore_permission = true;
-		if ( has_filter( 'edac_ignore_permission' ) ) {
-			$ignore_permission = apply_filters( 'edac_ignore_permission', $ignore_permission );
-		}
+		$ignore_permission = apply_filters( 'edac_ignore_permission', true );
 		foreach ( $rules as $rule ) {
 			// phpcs:ignore WordPress.DB.DirectDatabaseQuery.DirectQuery, WordPress.DB.DirectDatabaseQuery.NoCaching -- Using direct query for interacting with custom database, safe variable used for table name, caching not required for one time operation.
 			$results        = $wpdb->get_results( $wpdb->prepare( 'SELECT id, postid, object, ruletype, ignre, ignre_user, ignre_date, ignre_comment, ignre_global FROM %i where postid = %d and rule = %s and siteid = %d', $table_name, $postid, $rule['slug'], $siteid ), ARRAY_A );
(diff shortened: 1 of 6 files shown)


accessibility-checker.php
```diff
@@ -1656,12 +1651,10 @@ function edac_get_simplified_summary( $post = null ) {
...
-	$simplified_summary_heading = 'Simplified Summary';
-
-	// filter title.
-	if ( has_filter( 'edac_filter_simplified_summary_heading' ) ) {
-		$simplified_summary_heading = apply_filters( 'edac_filter_simplified_summary_heading', $simplified_summary_heading );
-	}
+	$simplified_summary_heading = apply_filters(
+		'edac_filter_simplified_summary_heading',
+		esc_html__( 'Simplified Summary', 'accessibility-checker' )
+	);
```
aristath, line 1657: Also made the text translatable

includes/classes/class-admin-notices.php
```diff
@@ -324,13 +324,18 @@ public function edac_review_notice_ajax() {
...
+			sprintf(
+				// translators: %s is the link to upgrade to pro, with "upgrade to pro" as the anchor text.
+				esc_html__( 'Whoops! It looks like your website is currently password protected. The free version of Accessibility Checker can only scan live websites. To scan this website for accessibility problems either remove the password protection or %s. Scan results may be stored from a previous scan.', 'accessibility-checker' ),
+				sprintf(
+					'<a href="https://equalizedigital.com/accessibility-checker/pricing/" target="_blank" aria-label="%1$s">%2$s</a>',
+					esc_attr__( 'Upgrade to accessibility checker pro. Opens in a new window.', 'accessibility-checker' ),
+					esc_html__( 'upgrade to pro', 'accessibility-checker' )
+				)
+			)
+		);
```
aristath, line 338: Also made sure the text for this notice is translatable and properly escaped.

includes/helper-functions.php
```diff
@@ -182,10 +182,7 @@ function edac_is_gutenberg_active() {
 	$gutenberg    = false;
 	$block_editor = false;
 
-	if ( has_filter( 'replace_editor', 'gutenberg_init' ) ) {
-		// Gutenberg is installed and activated.
-		$gutenberg = true;
-	}
+	$gutenberg = has_filter( 'replace_editor', 'gutenberg_init' );
```
aristath, line 185: Ideally we'd check if Gutenberg is active using a check like the one below:
```php
$gutenberg = defined( 'IS_GUTENBERG_PLUGIN' );
```

That's something we can do in a followup PR when we rething a bit how we check for plugins 👍

SteveJonesDev (approved):

PR #422, opened 2023-12-13
Improve code-quality/readability in the main file

aristath:
This PR contains minor tweaks in the main plugin file:
* Split long lines to multiple lines where appropriate
* Removed a couple of unnecessary variables
* Removed calls to `add_post_meta` before `update_post_meta`
* Simplified a couple of conditions
* Fixed some translations

### DIFF
--- a/accessibility-checker.php
+++ b/accessibility-checker.php
@@ -295,7 +295,7 @@ function edac_before_page_render() {
 		// Check the page if it hasn't already been checked.
 		global $post;
 		$checked = get_post_meta( $post->ID, '_edac_post_checked', true );
-		if ( false === boolval( $checked ) ) {
+		if ( ! $checked ) {
 			edac_validate( $post->ID, $post, $action = 'load' );
 		}
 	}
@@ -348,8 +348,17 @@ function edac_summary( $post_id ) {
 			$postid = $post_id;
 			$siteid = get_current_blog_id();
 
-			// phpcs:ignore WordPress.DB.DirectDatabaseQuery.DirectQuery, WordPress.DB.DirectDatabaseQuery.NoCaching  -- Using direct query for interacting with custom database, safe variable used for table name, caching not required for one time operation.
-			$rule_count = $wpdb->get_var( $wpdb->prepare( 'SELECT count(*) FROM %i where rule = %s and siteid = %d and postid = %d and ignre = %d', $table_name, $rule['slug'], $siteid, $postid, 0 ) );
+			// phpcs:ignore WordPress.DB.DirectDatabaseQuery.DirectQuery, WordPress.DB.DirectDatabaseQuery.NoCaching -- Using direct query for interacting with custom database, safe variable used for table name, caching not required for one time operation.
+			$rule_count = $wpdb->get_var(
+				$wpdb->prepare(
+					'SELECT count(*) FROM %i where rule = %s and siteid = %d and postid = %d and ignre = %d',
+					$table_name,
+					$rule['slug'],
+					$siteid,
+					$postid,
+					0
+				)
+			);
 
 			if ( ! $rule_count ) {
 				$rules_passed[] = $rule['slug'];
@@ -360,7 +369,16 @@ function edac_summary( $post_id ) {
 	$summary['passed_tests'] = round( count( $rules_passed ) / count( $rules ) * 100 );
 
 	// phpcs:ignore WordPress.DB.DirectDatabaseQuery.DirectQuery, WordPress.DB.DirectDatabaseQuery.NoCaching -- Using direct query for interacting with custom database, safe variable used for table name, caching not required for one time operation.
-	$summary['errors'] = intval( $wpdb->get_var( $wpdb->prepare( 'SELECT count(*) FROM %i where siteid = %d and postid = %d and ruletype = %s and ignre = %d', $table_name, get_current_blog_id(), $post_id, 'error', 0 ) ) );
+	$summary['errors'] = (int) $wpdb->get_var(
+		$wpdb->prepare(
+			'SELECT count(*) FROM %i where siteid = %d and postid = %d and ruletype = %s and ignre = %d',
+			$table_name,
+			get_current_blog_id(),
+			$post_id,
+			'error',
+			0
+		)
+	);
 
 	// count warnings.
 	$warnings_parameters = array( get_current_blog_id(), $post_id, 'warning', 0 );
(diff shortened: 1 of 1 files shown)


accessibility-checker.php
```diff
@@ -348,8 +348,17 @@ function edac_summary( $post_id ) {
...
+			$rule_count = $wpdb->get_var(
+				$wpdb->prepare(
+					'SELECT count(*) FROM %i where rule = %s and siteid = %d and postid = %d and ignre = %d',
+					$table_name,
+					$rule['slug'],
+					$siteid,
+					$postid,
+					0
+				)
+			);
```
aristath, line 361: The only change in queries was to split them to multiple lines so that it's easier to visually figure out the nesting levels and what the prepare does with the placeholders.

accessibility-checker.php
```diff
@@ -360,7 +369,16 @@ function edac_summary( $post_id ) {
...
+	$summary['errors'] = (int) $wpdb->get_var(
+		$wpdb->prepare(
+			'SELECT count(*) FROM %i where siteid = %d and postid = %d and ruletype = %s and ignre = %d',
+			$table_name,
+			get_current_blog_id(),
+			$post_id,
+			'error',
+			0
+		)
+	);
```
aristath, line 381: Split this one (and others below) to multiple lines, but also removed the `intval()` call, replacing it with an `(int)` typecase which is faster.

accessibility-checker.php
```diff
@@ -421,39 +458,25 @@ function edac_summary( $post_id ) {
 
 	$summary['content_grade'] = 0;
 	if ( class_exists( 'DaveChild\TextStatistics\TextStatistics' ) ) {
-		$text_statistics          = new DaveChild\TextStatistics\TextStatistics();
-		$summary['content_grade'] = floor( $text_statistics->fleschKincaidGradeLevel( $content ) );
+		$summary['content_grade'] = floor(
+			( new DaveChild\TextStatistics\TextStatistics() )->fleschKincaidGradeLevel( $content )
+		);
```
aristath, line 463: Just removed the `$text_statistics` variable here, it was not really necessary.

accessibility-checker.php
```diff
@@ -421,39 +458,25 @@ function edac_summary( $post_id ) {
...
-
-	if ( ! add_post_meta( $post_id, '_edac_summary_contrast_errors', $summary['contrast_errors'], true ) ) {
-		update_post_meta( $post_id, '_edac_summary_contrast_errors', $summary['contrast_errors'] );
-	}
+	update_post_meta( $post_id, '_edac_summary', $summary );
+	update_post_meta( $post_id, '_edac_summary_passed_tests', $summary['passed_tests'] );
+	update_post_meta( $post_id, '_edac_summary_errors', $summary['errors'] );
+	update_post_meta( $post_id, '_edac_summary_warnings', $summary['warnings'] );
+	update_post_meta( $post_id, '_edac_summary_ignored', $summary['ignored'] );
+	update_post_meta( $post_id, '_edac_summary_contrast_errors', $summary['contrast_errors'] );
```
aristath, line 479: `update_post_meta` calls `update_metadata`.
[Documentation for the `update_metadata` function](https://developer.wordpress.org/reference/functions/update_metadata/) mentions this:
> If no value already exists for the specified object ID and metadata key, the metadata will be added.

With that in mind, calling `add_post_meta` before `update_post_meta` is unnecessary.

The resulting code is cleaner, faster, and easier to digest.

accessibility-checker.php
```diff
@@ -467,9 +490,9 @@ function edac_anww_update_post_meta() {
 
 	$option_name = 'edac_anww_update_post_meta';
 
-	if ( get_option( $option_name ) === false && EDAC_ANWW_ACTIVE ) {
+	if ( ! get_option( $option_name ) && EDAC_ANWW_ACTIVE ) {
 		update_option( $option_name, true );
-	} elseif ( get_option( $option_name ) === true && ! EDAC_ANWW_ACTIVE ) {
+	} elseif ( get_option( $option_name ) && ! EDAC_ANWW_ACTIVE ) {
```
aristath, line 495: The `edac_anww_update_post_meta` option has a boolean value, so this condition was simplified

accessibility-checker.php
```diff
@@ -509,7 +540,22 @@ function edac_documentation_link( $rule ) {
...
+			'utm_campaign'     => 'wordpress-general',
+			'php_version'      => PHP_VERSION,
+			'platform'         => 'wordpress',
+			'platform_version' => $wp_version,
+			'software'         => 'free',
+			'software_version' => EDAC_VERSION,
+			'days_active'      => $days_active,
+		),
+		$rule['info_url']
+	);
```
aristath, line 558: Used `add_query_arg` here to build the URL instead of having a huge string with lots of concatenation.
`add_query_arg` is slightly slower, but in this case, the code becomes a lot easier to understand and maintain so it's worth the tradeoff.

accessibility-checker.php
```diff
@@ -525,11 +571,14 @@ function edac_output_simplified_summary( $content ) {
...
-		return $content . $simplified_summary;
+
+	if ( $simplified_summary ) {
+		if ( 'before' === $simplified_summary_position ) {
+			return $simplified_summary . $content;
+		}
+		if ( 'after' === $simplified_summary_position ) {
+			return $content . $simplified_summary;
+		}
 	}
```
aristath, line 582: Since both conditions depend on `$simplified_summary`, they were grouped/nested to make it clearer.

accessibility-checker.php
```diff
@@ -576,14 +628,29 @@ function edac_get_accessibility_statement() {
...
+		);
 	}
 
 	if ( $include_statement_link && $policy_page ) {
-		$statement .= esc_html__( 'Read our ', 'accessibility-checker' ) . '<a href="' . $policy_page . '">' . esc_html__( 'Accessibility Policy', 'accessibility-checker' ) . '</a>.';
+		$statement .= sprintf(
+			// translators: %1$s is a link to the accessibility policy page, with text "Accessibility Policy".
+			esc_html__( 'Read our %s', 'accessibility-checker' ),
+			'<a href="' . $policy_page . '">' . esc_html__( 'Accessibility Policy', 'accessibility-checker' ) . '</a>.'
+		);
```
aristath, line 653: Fixed concatenation in translation strings

SteveJonesDev (approved):

PR #429, opened 2023-12-18
:test_tube: :alembic: :construction:  Proof of concept to abstract the rules logic :construction: 

aristath:
This PR is just a prototype to investigate how we could abstract the logic for rules in a  more object-oriented way.

Please note that there are some modifications where we include the files and run the evaluations to allow running both the old and new implementations for the sake of this experiment/POC, showcasing how we could do it. I'm also hard-including the new classes because I'm treating this as a rough proof of concept to demonstrate an idea :sweat_smile: 

What is your opinion on doing something like this @boonedev @jdevalk @SteveJonesDev ?

Right now we have some basic data for rules in the `includes/rules.php` file, and then we have the rule's evaluation in `includes/rules/{RULE_SLUG}.php` files. This PR attempts to do the following:
* Adds a `Rules` object, which can act as a collection of rules.
* Adds an abstract `Rule` object, which individual rules can extend
* Adds a `Rule_Aria_Hidden` as a demostration of how a rule can be implemented, extending the `Rule` object. This class unifies the rule info from `includes/rules.php` and `includes/rules/aria_hidden.php`
* Deletes the `includes/rules/aria_hidden.php` file since the info is now in the `Rule_Aria_Hidden` object, and removes the rule from `includes/rules.php`


accessibility-checker.php
```diff
@@ -129,6 +129,30 @@
...
+				'info_url'  => $rule->get_info_url(),
+				'slug'      => $rule->get_slug(),
+				'rule_type' => $rule->get_type(),
+				'summary'   => $rule->get_summary(),
+			);
+		}
+		return $rules;
+	}
+);
+
```
aristath, line 155: This is just for the POC, to show how we can add rules in a manner that is backwards-compatible.

accessibility-checker.php
```diff
@@ -231,8 +255,12 @@ function edac_include_rules_files() {
 	}
 	foreach ( $rules as $rule ) {
 		if ( ( array_key_exists( 'ruleset', $rule ) && 'php' === $rule['ruleset'] )
-			|| ( ! array_key_exists( 'ruleset', $rule ) && $rule['slug'] ) 
+			|| ( ! array_key_exists( 'ruleset', $rule ) && $rule['slug'] )
 		) {
+			$rule_file_path = plugin_dir_path( __FILE__ ) . 'includes/rules/' . $rule['slug'] . '.php';
+			if ( ! file_exists( $rule_file_path ) ) {
+				continue;
+			}
```
aristath, line 263: This was done simply because I removed the `aria_hidden` rule file

amberhinds:
@aristath have you looked at how the rules are changing with the transition to JavaScript scanning? https://github.com/equalizedigital/accessibility-checker/tree/feature/javascript-scan

aristath:
@amberhinds thank you, I had not seen that!
So... if we do JS checks, what's the plan for checks running in the editor? :thinking:

boonedev:
Hi @aristath, thanks for the code cleanup help! Here's an overview of how checks are running in the editor:
https://github.com/equalizedigital/accessibility-checker-pro/blob/develop/README_DEV.txt

SteveJonesDev:
@aristath, see the readme @boonedev posted for details. Currently, the only rule converted to Javascript is the color contrast as it was the rule causing the most false positives. We'll have a hybrid approach until all rules are converted over using either axe-core or a custom Javascript rule. So we'll still need to refactor the rules logic as you've demonstrated. Unless we push forward and convert all the rules now.

SteveJonesDev:
All rules have been converted to JavaScript. Closing as no more relevant.

PR #431, opened 2023-12-18
Fixes #430

### LINKED ISSUE #430: Highlight panel blocks click interactions when closed
The in-page checker, when minimized, is a small yellow button at the lower right.

The wrapper around the button in this state is transparent, but is still 400px by 50px div, and is z-indexed in front of the content:

```
.edac-highlight-panel {
  bottom: 15px !important;
  max-width: 400px !important;
  position: fixed !important;
  right: 15px !important;
  width: 400px !important;
  z-index: 2147483647 !important;
}
```
To fix: set max-width to equal the button width when the panel has not yet been opened.

See: https://wordpress.org/support/topic/highlight-panel-blocks-click-interactions-when-closed/

boonedev:
this PR sets the panel width to 400px only when it is visible

### DIFF
--- a/includes/classes/class-rest-api.php
+++ b/includes/classes/class-rest-api.php
@@ -15,272 +15,272 @@ use EDAC\Helpers;
 /**
  * Class that initializes and handles the REST api
  */
-class REST_Api
-{
-
-    /**
-     * If class has already been initialized.
-     *
-     * @var boolean
-     */
-    private static $initialized = false;
-
-    /**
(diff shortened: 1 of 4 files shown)


SteveJonesDev (changes requested):
Please fix CS issues

SteveJonesDev (approved):
Nevermind, I see you've fixed the CS issues.

SteveJonesDev:
Good to merge

PR #435, opened 2023-12-20
The `edac_log` function does not exist

aristath:
I noticed a couple of calls to an `edac_log` function. However, that function does not exist anywhere.
I found `edacp_log` - and I _assume_ that is the one we wanted to use, so I renamed the calls to the right function.


SteveJonesDev (approved):
I removed the edac_log function as it was flagging CS issues. I thought I removed all this EDAC_DEBUG logic, but I guess not. We can use edacp_log for now. We could also just use `error_log()`. I'll merge this for now and we can decide later.

PR #452, opened 2024-01-11
396 properly enqueue scripts and styles

boonedev:
This PR:
 - Removes enqueue-scripts.php and replaces it with ECAD\Admin\Enqueue_Admin and EDAC\Inc\Enqueue_Frontend
 - Adds maybe_enqueue_email_opt_in_script() to EDAC/Admin/Enqueue_Admin
 - Adds js & css for emailOptIn to webpack build 
 - Removes zoho form from Edac\Admin\Welcome_Page and replaces with it with ActiveCampaign


accessibility-checker.php
```diff
@@ -134,9 +133,8 @@
 /**
  * Filters and Actions
  */
-add_action( 'admin_enqueue_scripts', 'edac_admin_enqueue_scripts' );
-add_action( 'admin_enqueue_scripts', 'edac_admin_enqueue_styles' );
-add_action( 'wp_enqueue_scripts', 'edac_enqueue_scripts' );
+add_action( 'admin_enqueue_scripts', array( 'EDAC\Admin\Enqueue_Admin', 'enqueue' ) );
```
SteveJonesDev, line 136: Shouldn't this be moved to the `class-admin.php` so it's never executed unless in the admin, consistency, and to keep the main plugin file clean?

accessibility-checker.php
```diff
@@ -134,9 +133,8 @@
...
+add_action( 'wp_enqueue_scripts', array( 'EDAC\Inc\Enqueue_Frontend', 'enqueue' ) );
```
SteveJonesDev, line 137: Shouldn't this be moved to the `class-plugin.php` so it's never executed unless on the front-end, consistency, and to keep the main plugin file clean?

SteveJonesDev (changes requested):
Just a couple of changes.

accessibility-checker.php
boonedev: @SteveJonesDev Good catch, thanks.

SteveJonesDev (approved):

PR #453, opened 2024-01-12
Hotfix - basic auth fix

boonedev:
This PR:

- Bugfix: changes misnamed js variable from edacp_full_site_scan_app to edacpFullSiteScanApp
- Adds EDAC\Helpers\is_basic_auth()
- replaces get_option( 'edac_password_protected') checks with EDAC\Helpers\is_basic_auth() 
- Adds basicAuth to js edac_editor_app
- Adds auth warning to editor page when user saves a post in free on a site that requires auth

### DIFF
--- a/accessibility-checker.php
+++ b/accessibility-checker.php
@@ -835,7 +835,7 @@ function edac_summary_ajax() {
 
 	// password check.
 	if (
-		!(
+		! (
 			EDAC_KEY_VALID === true &&
 			edac_check_plugin_active( 'accessibility-checker-pro/accessibility-checker-pro.php' )
 		) &&
--- a/accessibility-checker.php
+++ b/accessibility-checker.php
@@ -834,7 +834,13 @@ function edac_summary_ajax() {
 	$html['content'] = '';
 
 	// password check.
-	if ( boolval( get_option( 'edac_password_protected' ) ) === true ) {
+	if (
+		!(
+			EDAC_KEY_VALID === true &&
+			edac_check_plugin_active( 'accessibility-checker-pro/accessibility-checker-pro.php' )
+		) &&
+		\EDAC\Helpers::is_basic_auth() 
+	) {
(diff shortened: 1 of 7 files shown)


accessibility-checker.php
```diff
@@ -834,7 +834,13 @@ function edac_summary_ajax() {
...
 
 	// password check.
-	if ( boolval( get_option( 'edac_password_protected' ) ) === true ) {
+	if (
+		! (
+			EDAC_KEY_VALID === true &&
+			edac_check_plugin_active( 'accessibility-checker-pro/accessibility-checker-pro.php' )
+		) &&
+		\EDAC\Helpers::is_basic_auth() 
+	) {
```
boonedev, line 843: Only show the "Whoops! It looks like your website is currently password protected..." message when editing pages if Pro is not valid/active and we've tested that basic auth is in place.

src/editorApp/checkPage.js
```diff
@@ -5,11 +5,12 @@ import { showNotice } from './../common/helpers';
 
 const API_URL = edac_editor_app.edacApiUrl;
 
+
 let HEADERS;
-if (typeof (edacp_full_site_scan_app) === 'undefined') {
+if (typeof (edacpFullSiteScanApp) === 'undefined') {
```
boonedev, line 10: This misnamed variable is the root cause of the issue b/c the user/pass were not being passed so auth failed.

src/editorApp/index.js
```diff
@@ -1,102 +1,73 @@
...
-			  
-				  return response.status;
-				} catch (error) {
-				  return 401; // Default status for error
-				}
-			};
-	
-
-			checkApi().then((status) => {
-
```
boonedev, line 42: We're using serverside to test if auth is in place, so no need for this.

SteveJonesDev (approved):

PR #456, opened 2024-01-19
Auth issue fix

boonedev:
This PR fixes https://github.com/equalizedigital/accessibility-checker-pro/issues/186
 - Removes unreliable server side check for determining if site is password protected
 - Removes js-based password protected warnings
 - Fixes PHP Deprecated:  Automatic conversion of false to array is deprecated in /var/www/html/wp-content/plugins/accessibility-checker/includes/classes/class-scans-stats.php on line 440

### DIFF
--- a/includes/classes/class-helpers.php
+++ b/includes/classes/class-helpers.php
@@ -207,41 +207,4 @@ class Helpers {
 	
 		return false;
 	}
-
-	/**
-	 * Determine if this site is using basic auth.
-	 *
-	 * @return boolean
-	 */
-	public static function is_basic_auth() {
-		
-		$key = 'edac_auth_type';
-
-		$status = get_transient( $key );
-
-		$status = false;
-
-		if ( false === $status ) {
-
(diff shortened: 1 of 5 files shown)


includes/classes/class-helpers.php
```diff
@@ -207,41 +207,4 @@ public static function is_domain_loopback( $domain ) {
...
-			set_transient( $key, $status, 30 );
-
-		}
-
-		if ( 'basic' === $status ) {
-			return true;
-		}
-
-		return false;
-	}
```
boonedev, line 246: This is not reliable (slow and can fail in loopback).

src/editorApp/index.js
```diff
@@ -8,63 +8,21 @@ window.addEventListener('DOMContentLoaded', () => {
...
-					}
-
-				});
-
-			} else {
-				debug("Gutenberg is not enabled.");
-			}
-
-
-
```
boonedev, line 64: Unnecessary.

SteveJonesDev (approved):

PR #477, opened 2024-01-26
461 scans status php deprecated

boonedev:
This PR fixes an incorrectly named variable.

### DIFF
--- a/admin/class-scans-stats.php
+++ b/admin/class-scans-stats.php
@@ -13,6 +13,7 @@ namespace EDAC\Admin;
 class Scans_Stats {
 
 
+
 	/**
 	 * Number of seconds to return results from cache.
 	 *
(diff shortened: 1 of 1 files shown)


admin/class-scans-stats.php
```diff
@@ -13,6 +13,7 @@
 class Scans_Stats {
 
 
+
```
aristath, line 16:
```suggestion
```

admin/class-scans-stats.php
```diff
@@ -398,7 +399,7 @@ public function issues_summary_by_post_type( $post_type ) {
 		$data['cache_id']   = $transient_name;
 		$data['cached_at']  = time();
 		$data['expires_at'] = time() + $this->cache_time;
-		$cache['cache_hit'] = false;
+		$data['cache_hit']  = false;
```
aristath, line 402: Good catch! :+1:

aristath (approved):

PR #489, opened 2024-02-06
🚧 402 use fewer postmeta values 🚧

boonedev:
This PR adds a standard way to handle the edac specific data we stored about a site and posts.
1) Adds a singleton Options class:
 - Uses a single interface to set/get/delete the data items. Consolidatse the separate wp_options records into a single wp_options record that stores the grouped data as an array.
- Enforces rules to insure data is of the name, datatype and value we expect.
- Sets the default values for items.
- Performs a one-time migration of the legacy data items to the new format.
- Handles backward compatibility when user is running an older version of Pro or Audit that directly accesses options with - get_option/update_option/delete_option.
 - Adds EDACAdminOptionsTest.php unit test.

2)  Adds a Post_Options class:
 - Uses a single interface to set/get/delete the data items.
 - Enforces rules to insure data is stored in the format we expect (grouped into a single record or as separate records).
 - Enforces rules to insure data is of the name, datatype and value we expect.
 - Sets the default values for grouped items.
 - Handles backward compatibility for legacy _edac_summary data.
 - Handles backward compatibility when user is running an older version of Pro or Audit that directly accesses options with get_post_metadata/update_post_metadata/delete_post_metadata. 
- Adds EDACAdminPostOptionsTest.php unit test.


accessibility-checker.php
```diff
@@ -187,12 +189,12 @@ function edac_update_database() {
 		require_once ABSPATH . 'wp-admin/includes/upgrade.php';
 		dbDelta( $sql );
 
-	}
+		// Update database version option.
```
boonedev, line 192: Moved this to inside the if statement so it will only make the db call if the version has changed.

Issue #464, opened 2024-01-25
Convert edac_insert_rule_data function to a class


SteveJonesDev:
https://github.com/equalizedigital/accessibility-checker/blob/develop/includes/insert.php

Should be an `EDAC\Admin` class.

PR #513, opened 2024-02-23
Move edac_insert_rule_data to an inserter class

pattonwebz:
Moves the `edac_insert_rule_data` function to a new `Insert_Rule_Data` class with a single method named `insert` that takes the same passed params.

- I added type hints on the parameters (should `post` actually be a `\WP_Post` type rather than an `object`?).
- Added an early bail if one of the required params aren't passed (typehints technically prevent this happening but having it makes it more clear what is going on).
- Started adding tests (which still needs more work to cover more of the method).

I have some thoughts about potential refactor of this function so that it can have return values that are likely to give more certainty about what has happened. My initial thought was `bool|int` and never `void`. With true being success, false being fail and the `int` meaning it was a duplicate but would also evaluate to a truth value for checking that might be added later after it's called.

Closes: #464


SteveJonesDev:
@pattonwebz, are you still working on this PR since it's in draft mode?

pattonwebz:
> @pattonwebz, are you still working on this PR since it's in draft mode?

I still need to work on the tests here for better coverage. If you want to review the actual code changes, that'd be great.

I plan to revisit the tests here tomorrow and the other issues assigned to me.

admin/class-insert-rule-data.php
```diff
@@ -0,0 +1,113 @@
...
+			'rule'          => $rule,
+			'ruletype'      => $ruletype,
+			'object'        => esc_attr( $rule_obj ),
+			'recordcheck'   => 1,
+			'user'          => get_current_user_id(),
+			'ignre'         => 0,
+			'ignre_user'    => null,
+			'ignre_date'    => null,
+			'ignre_comment' => null,
+			'ignre_global'  => 0,
```
pattonwebz, line 48: Why are these `ignre` and not `ignore`?

SteveJonesDev: It's been four years but I believe I named it that way out of an abundance of caught to prevent conflicts or reserved keywords.

pattonwebz: That makes sense, I figured there was some reason behind it that wasn't just a simple typo haha

Hard to change it at this point if it was just a typo though :rofl:

tests/phpunit/InsertRuleDataTest.php
```diff
@@ -0,0 +1,96 @@
...
+	/**
+	 * Tests the insert method would return expected data types.
+	 */
+	public function testRuleInserterReturnLogic() {
+		$post     = $this->factory()->post->create_and_get();
+		$rule     = 'rule';
+		$ruletype = 'ruletype';
+		$rule_obj = 'rule_obj';
+
+		sleep( 2 );
```
pattonwebz, line 67: This was added only so I had 2 seconds to validate that the tables existed. It should have been removed before pushing

```suggestion
```

admin/class-insert-rule-data.php
```diff
@@ -0,0 +1,119 @@
...
+				);
+
+			}
+		}
+
+		// Insert new records.
+		if ( ! $results ) {
+
+			// filter post types.
+			$rule_data = apply_filters( 'edac_filter_insert_rule_data', $rule_data );
```
SteveJonesDev, line 109: @pattonwebz, should we sanitize this data to ensure it's been filtered in a safe manner?

Example:
```php
// Apply filters to allow modification of the rule data.
$rule_data = apply_filters( 'edac_filter_insert_rule_data', $rule_data );

// Sanitize the filtered data.
$rule_data_sanitized = array(
    'postid'        => absint( $rule_data['postid'] ),
    'siteid'        => absint( $rule_data['siteid'] ),
    'type'          => sanitize_text_field( $rule_data['type'] ),
    'rule'          => sanitize_text_field( $rule_data['rule'] ),
    'ruletype'      => sanitize_text_field( $rule_data['ruletype'] ),
    'object'        => sanitize_text_field( $rule_data['object'] ),
    'recordcheck'   => absint( $rule_data['recordcheck'] ),
    'user'          => absint( $rule_data['user'] ),
    'ignore'        => absint( $rule_data['ignore'] ), // Note: Fixed typo 'ignre' to 'ignore'
    'ignore_user'   => isset( $rule_data['ignore_user'] ) ? absint( $rule_data['ignore_user'] ) : null,
    'ignore_date'   => isset( $rule_data['ignore_date'] ) ? sanitize_text_field( $rule_data['ignore_date'] ) : null,
    'ignore_comment'=> isset( $rule_data['ignore_comment'] ) ? sanitize_text_field( $rule_data['ignore_comment'] ) : null,
    'ignore_global' => absint( $rule_data['ignore_global'] ),
);

// Now, $rule_data_sanitized contains the sanitized data ready for insertion.
```

admin/class-insert-rule-data.php
```diff
@@ -0,0 +1,119 @@
...
+	 * @param object $post     The post object.
+	 * @param string $rule     The rule.
+	 * @param string $ruletype The rule type.
+	 * @param string $rule_obj The object.
+	 *
+	 * @return void|int
+	 */
+	public function insert( object $post, string $rule, string $ruletype, string $rule_obj ) {
+
+		if ( ! $post || ! $rule || ! $ruletype || ! $rule_obj ) {
```
SteveJonesDev, line 33: Since these are typed parameters these checks can be more robust. `empty()` checks for `0`, `null`, `false`, an empty string `''`, and an unset variable, covering more cases where the variable might not be useful.

Example:
```php
if ( empty( $post->ID ) || empty( $rule ) || empty( $ruletype ) || empty( $rule_obj ) ) {
    return;
}
```

admin/class-insert-rule-data.php
```diff
@@ -0,0 +1,119 @@
...
+	 * Insert rule data into database
+	 *
+	 * @since 2.0.0
+	 *
+	 * @param object $post     The post object.
+	 * @param string $rule     The rule.
+	 * @param string $ruletype The rule type.
+	 * @param string $rule_obj The object.
+	 *
+	 * @return void|int
```
SteveJonesDev, line 29: Update this to:
` @return void|int The ID of the newly inserted record, or void if no new record is inserted.`

admin/class-insert-rule-data.php
```diff
@@ -0,0 +1,119 @@
...
+ * @since 2.0.0
+ */
+class Insert_Rule_Data {
+
+	/**
+	 * Insert rule data into database
+	 *
+	 * @since 2.0.0
+	 *
+	 * @param object $post     The post object.
```
SteveJonesDev, line 24: Update this to:
`@param object $post     The post object. Must have a valid ID property.`

SteveJonesDev (changes requested):

admin/class-insert-rule-data.php
```diff
@@ -0,0 +1,119 @@
+<?php
+/**
+ * Inserts rule data about a post to the database
+ *
+ * @since 2.0.0
```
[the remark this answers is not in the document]
pattonwebz: I'll swap this to `1.10.0` and do the same in any other PRs, so there's no need to flag all the places in the others; I'll sort them out.

admin/class-insert-rule-data.php
pattonwebz: I see that you fixed the `ignre` to `ignore,` but we can't do that without first remapping the columns in the table. Would you like me to make an issue to tackle that in the future?

includes/deprecated/deprecated.php
```diff
@@ -73,3 +74,19 @@ function edac_delete_cpt_posts( $post_type ) {
...
+ * @deprecated 1.10.0
+ *
+ * @param object $post     The post object.
+ * @param string $rule     The rule.
+ * @param string $ruletype The rule type.
+ * @param string $rule_obj The object.
+ * @return void|int
+ */
+function edac_insert_rule_data( $post, $rule, $ruletype, $rule_obj ) {
+	_deprecated_function( __FUNCTION__, '2.0.0', 'EDAC\Admin\Insert_Rule_Data' );
```
SteveJonesDev, line 90: Update version to 1.10.0

SteveJonesDev (changes requested):

SteveJonesDev (approved):
Good to merge.

Issue #510, opened 2024-02-21
Load Notyf styles locally and remove CDN call to stylesheet


SteveJonesDev:
https://github.com/equalizedigital/accessibility-checker/blob/develop/src/common/helpers.js#L40

PR #520, opened 2024-02-28
Don't load notyf styles from CDN

pattonwebz:
Refactors the inclusion of notyf styles to pull from the already installed notyf package and have it bundled into a local CSS file for inclusion.

Fixes: #510


SteveJonesDev (approved):
Looks good. Please merge.

Issue #448, opened 2024-01-02
Make aria-hidden warning smarter


amberhinds:
## Is your feature request related to a problem? Please describe.
Currently, we flag an aria-hidden warning on any element that has `aria-hidden="true"` on it and people then have to go in and assess the HTML that surrounds that element to see if aria-hidden was correctly used. This can result in extra noise and work that might be overwhelming for people who don't know how to read HTML.

## Describe the solution you'd like
If possible, I would like to expand this warning to review any link or button tags that contain aria-hidden elements and assess if there is either screen reader text, visible text, or an aria-label present. If so, then an aria-hidden warning would not be flagged.

## Example of passing code that would not flag a warning

```
<!--Button with aria-label-->
<button type="button" aria-haspopup="true" aria-label="Open menu" class="components-button wp-block-navigation__responsive-container-open" inert="true">
    <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" width="24" height="24" aria-hidden="true" focusable="false"><rect x="4" y="7.5" width="16" height="1.5"></rect><rect x="4" y="15" width="16" height="1.5"></rect></svg>
</button>

<!--Button with screen reader text-->
<button type="button" aria-haspopup="true" class="components-button wp-block-navigation__responsive-container-open" inert="true">
    <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" width="24" height="24" aria-hidden="true" focusable="false"><rect x="4" y="7.5" width="16" height="1.5"></rect><rect x="4" y="15" width="16" height="1.5"></rect></svg>
<span class="sr-only">Open menu</span>
</button>

<!--Button with visible text-->
<button type="button" aria-haspopup="true" class="components-button wp-block-navigation__responsive-container-open" inert="true">
    <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" width="24" height="24" aria-hidden="true" focusable="false"><rect x="4" y="7.5" width="16" height="1.5"></rect><rect x="4" y="15" width="16" height="1.5"></rect></svg>
Menu
</button>

<!--Link with aria-label-->
<a href="/about" aria-label="About Us">
    <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" width="24" height="24" aria-hidden="true" focusable="false"><rect x="4" y="7.5" width="16" height="1.5"></rect><rect x="4" y="15" width="16" height="1.5"></rect></svg>
</a>

[...]

PR #523, opened 2024-03-07
Improve aria-hidden warning rule to cover some valid use cases for adding it to elements

pattonwebz:
Right now, the aria-hidden rule flags a warning for every instance that it detects and tasks users with looking at each one and its context to determine whether it is valid. If they find it valid, they are instructed to ignore the issue.

This PR seeks to add some smarts to the checker so that it can try not to flag some instances as warnings when we can tell they are valid already.

It was pretty easy to detect items with labels on the parent, but screen reader text and visible text are a little more challenging and are still being worked through.

* Unit tests have been added to try to cover all the new functionality

Closes: #448

### DIFF
--- a/includes/rules/aria_hidden.php
+++ b/includes/rules/aria_hidden.php
@@ -32,14 +32,13 @@ function edac_rule_aria_hidden( $content, $post ) { // phpcs:ignore -- $post is 
 				}
 			}
 
-			$parent_node             = $element->parent();
-			$tags_for_further_checks = array(
-				'button',
-				'a',
-			);
+			$parent_node = $element->parent();
 			if (
 				$parent_node &&
-				in_array( strtolower( $parent_node->tag ), $tags_for_further_checks, true )
+				(
+					strtolower( $parent_node->tag ) === 'button' ||
+					strtolower( $parent_node->tag ) === 'a'
+				)
 			) {
 				if ( edac_rule_aria_hidden_valid_parentnode_condition_check( $parent_node, $element ) ) {
 					continue;
--- a/includes/rules/aria_hidden.php
+++ b/includes/rules/aria_hidden.php
@@ -32,8 +32,15 @@ function edac_rule_aria_hidden( $content, $post ) { // phpcs:ignore -- $post is 
 				}
 			}
 
-			$parent_node = $element->parent();
-			if ( $parent_node && ( strtolower( $parent_node->tag ) === 'button' || strtolower( $parent_node->tag ) === 'a' ) ) {
+			$parent_node             = $element->parent();
+			$tags_for_further_checks = array(
+				'button',
+				'a',
+			);
+			if (
+				$parent_node &&
+				in_array( strtolower( $parent_node->tag ), $tags_for_further_checks, true )
+			) {
 				if ( edac_rule_aria_hidden_valid_parentnode_condition_check( $parent_node, $element ) ) {
 					continue;
 				}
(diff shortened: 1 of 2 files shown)


SteveJonesDev (changes requested):
See comment on issue: https://github.com/equalizedigital/accessibility-checker/issues/448#issuecomment-1998517248

includes/rules/aria_hidden.php
```diff
@@ -18,11 +18,44 @@ function edac_rule_aria_hidden( $content, $post ) { // phpcs:ignore -- $post is
...
-			if ( stristr( $element->getAttribute( 'class' ), 'wp-block-spacer' ) ) {
-				continue;
+			foreach ( $attributes_that_make_this_valid_hidden_use as $attribute => $value ) {
+				if ( stristr( $element->getAttribute( $attribute ), $value ) ) {
+					continue 2;
+				}
+			}
+
+			$parent_node = $element->parent();
+			if ( $parent_node && ( strtolower( $parent_node->tag ) === 'button' || strtolower( $parent_node->tag ) === 'a' ) ) {
```
SteveJonesDev, line 36: Improve readability

```php
			if ( 
				$parent_node && 
				( 
					strtolower( $parent_node->tag ) === 'button' || 
					strtolower( $parent_node->tag ) === 'a' 
				) 
			) {
```

includes/rules/aria_hidden.php
```diff
@@ -31,3 +64,76 @@ function edac_rule_aria_hidden( $content, $post ) { // phpcs:ignore -- $post is
...
+ * @since 1.10.0
+ *
+ * @param object $parent_node  The parent element.
+ * @param object $element The element.
+ *
+ * @return bool
+ */
+function edac_rule_aria_hidden_valid_parentnode_condition_check( object $parent_node, object $element ): bool {
+	// bail early if we don't have a parent node or an element node.
+	if ( ! $parent_node || ! $element ) {
```
SteveJonesDev, line 112: Readability:
```php
	if ( 
		! $parent_node || 
		! $element 
	) {
```

SteveJonesDev (changes requested):
The inner text change works nicely. Just a couple readability changes and this should be good to merge.

includes/rules/aria_hidden.php
pattonwebz: Seeing it laid out like this makes me think that I should reduce the number of lines in the condition. What do you think of this alternative?

```
			$parent_node             = $element->parent();
			$tags_for_further_checks = array(
				'button',
				'a',
			);
			if (
				$parent_node &&
				in_array( strtolower($parent_node->tag), $tags_for_further_checks, true )
			) {
```

SteveJonesDev: I'd rather it be a direct comparison than taking on the additional overhead of the `in_array` since the array isn't likely to change. It's a marginal performance improvement but it can add up throughout the code base. I was tempted to ask you to make the `$attributes_that_make_this_valid_hidden_use` a direct comparison as well be I think that one is okay for now.

I know code readability is subjective but I find this more readable anyways.

```php
			if ( 
				$parent_node && 
				( 
					strtolower( $parent_node->tag ) === 'button' || 
					strtolower( $parent_node->tag ) === 'a' 
				) 
			) {
```

pattonwebz: @SteveJonesDev gotcha, I updated this to align with your suggestion now. LMK if there was any other feedback.

SteveJonesDev (approved):
Good to merge.

PR #525, opened 2024-03-08
Solve some unexpected situations with frontend highlighter when rules are filtered off

### LINKED ISSUE #524: "Null" issues reported for checks that are being disabled by a filter
### Please give us a description of what happened.

When a user has already run a scan and then adds a filter that disables a check, positives on disabled checks can still show up in our reports as a "null" value.

### Please describe what you expected to happen and why.

The expected result of disabling a check is that the disabled check would be filtered out everywhere, including in scan results.

### How can we reproduce this behavior?
1. Run a scan on a page where a specific check will trigger a positive result.
2. Disable the triggered check by adding a filter to the theme or functions file.
3. Go back to the report on that page and you should see a "null" value displaying in front-end and potentially other areas of our reports.

### Technical info

N/A

* Which browser is affected (or browsers): 
- [x] Internet Explorer
- [x] Edge
- [x] Chrome
- [x] Firefox
- [x] Safari
- [x] Brave

* Relevant plugins in case of a bug:

This is not a result of a plugin conflict.

* Tested with theme:

N/A

pattonwebz:
When someone filters out a certain rule after a page is scanned, the Ajax call still returns those rules to the highlighter in a 'null' state without any information. This PR seeks to not return null data to the highlighter so they are not shown on the frontend when inspecting the page issues.

* Prevent filtered out rule issues being passed down to the highlighter.
* Also disable the buttons in the highlighter until we know we have issues to display.

<details>
  <summary>Widget when loading doesn't show nav buttons anymore</summary>
  
Screenshot from 2024-03-08 21-35-49

</details>

<details>
  <summary>Show buttons only after we know there are issues to display</summary>
  
Screenshot from 2024-03-08 21-36-08

</details>

<details>
  <summary>Show no buttons when no issues were returned from ajax call</summary>

Screenshot from 2024-03-08 21-36-55

</details>

Fixes #524

### DIFF
--- a/accessibility-checker.php
+++ b/accessibility-checker.php
@@ -10,7 +10,7 @@
  * Plugin Name:       Accessibility Checker
  * Plugin URI:        https://a11ychecker.com
  * Description:       Audit and check your website for accessibility before you hit publish. In-post accessibility scanner and guidance.
- * Version:           1.9.1
+ * Version:           1.9.2
  * Author:            Equalize Digital
  * Author URI:        https://equalizedigital.com
  * License:           GPL-2.0+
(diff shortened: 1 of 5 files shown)


SteveJonesDev (approved):
Looks good @pattonwebz! I will update the version and readme file, merge, and release this minor version.

SteveJonesDev (approved):

Issue #511, opened 2024-02-22
Missing_Transcript rule falsely flags URL strings in hyperlinks as errors


christophermhinds:
In your missing transcript check you look for the following file extensions 3gp, aif, aiff, asf, asx, avi, flv, m4a, m4p, mov, mp2, mp3, mp4, mpa, mpeg, mpeg2, mpg, mpv, ogg, ogv, qtl, ra, ram, smi, smil, wav, wax, webm, wma, wmp, or wmx. However, I've come across a number of false positives for this error when a link to a specific website has been used, for example https://www.wma.net/ https://www.mpdl.mpg.de/en/ Attached is an image of the error report. I think it might be worthwhile hardening up this check slightly

**Related ticket:** https://equalizedigital.zendesk.com/agent/tickets/6245

![Screenshot-2024-02-22-113217](https://github.com/equalizedigital/accessibility-checker/assets/25856499/da1367c8-a8d9-4d14-9834-c5441ada85c8)

PR #530, opened 2024-03-12
Tweak missing_transcript rule to try to avoid some false positives

pattonwebz:
The missing transcript rule can detect a false positive when the URL of a site or domain contains one of the extensions it tries to match against. Eg: www.wma.com

This PR attempts to avoid those by shifting the check to match against only the EXTENSION of href values and not matching within the value. That might not be the right approach, but pushing this up as an early attempt will add more tests to validate it suits and doesn't break any existing content checking that should pass (or allow some that should fail).

Fixes: #511


SteveJonesDev (approved):
Good to merge.

Issue #531, opened 2024-03-13
Subscriber restrictions needed in the WP admin


amberhinds:
* [x] I've read and understood the [contribution guidelines](https://github.com/equalizedigital/accessibility-checker/blob/develop/.github/CONTRIBUTING.md).
* [x] I've searched for any related issues and avoided creating a duplicate issue.

### Please give us a description of what happened.
I'm logged in as a subscriber, and I can see the accessibility status of the website:
<img width="1752" alt="Screenshot 2024-03-12 at 10 40 31 PM" src="https://github.com/equalizedigital/accessibility-checker/assets/13153456/a563239f-ace2-47f3-bd58-2a562d452d42">

Only the main page is visible, not sub-pages, but this should not be visible at all.

Also the dashboard widget should not be visible
<img width="560" alt="Screenshot 2024-03-12 at 10 42 22 PM" src="https://github.com/equalizedigital/accessibility-checker/assets/13153456/7e4a5b8f-af22-4955-b288-46d42ca614b0">

### How can we reproduce this behavior?
Log in as a subscribe on [one of our testing sites](https://demosite.equalizedigital.com) 

### Technical info
* WordPress version: 6.4.3
* Accessibility Checker version: 1.6.0

amberhinds:
I'm calling this critical because it could leak really sensitive information about websites.

pattonwebz:
This was resolved with #533 and released in v1.9.3

PR #532, opened 2024-03-13
Add some capability checking before adding the welcome page, outputting dashboard widget or notices

pattonwebz:
To prevent subscriber level roles accessing the welcome page or seeing the dashboard widget/notices some basic capability checks are added to gate it.

* Capability check of 'edit_posts' for the welcome page (aka contributor)
* Capability check of 'edit_posts' for widget and notices (aka contributor) - but this one runs through a new filter so that this can be edited by users if they choose.
* Notices are now wrapped in a new method that runs one `init` so that the capability checks can function after current_user is set.

Fixes: #531


pattonwebz:
Closing this to base from main as a hotfix

PR #533, opened 2024-03-13
Add a capability check for 'edit_posts' before outputting dashboard w…

pattonwebz:
To prevent subscriber-level roles from accessing the welcome page or seeing the dashboard widget/notices, some basic capability checks are added to gate it.

* Capability check of 'edit_posts' for the welcome page (aka contributor)
* Capability check of 'edit_posts' for widget and notices (aka contributor) - but this one runs through a new filter so that users can edit this if they choose.
* Notices are now wrapped in a new method that runs one `init` so that the capability checks can function after current_user is set.

Fixes: #531

### DIFF
--- a/admin/class-helpers.php
+++ b/admin/class-helpers.php
@@ -218,7 +218,7 @@ class Helpers {
 	 *
 	 * @since 1.9.3
 	 */
-	public static function current_user_can_see_widget(): bool {
+	public static function current_user_can_see_widgets_and_notices(): bool {
 		/**
 		 * Filter the capability required to view the dashboard widget.
 		 *
--- a/admin/class-helpers.php
+++ b/admin/class-helpers.php
@@ -226,9 +226,6 @@ class Helpers {
 		 *
 		 * @param string $capability The capability required to view the dashboard widget.
 		 */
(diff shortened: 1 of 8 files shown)


admin/class-helpers.php
```diff
@@ -211,4 +211,21 @@ function ( $rule ) {
 		}
 		return $results;
 	}
+
+	/**
+	 * Do a capability check for the current user to ensure they have the required capability
+	 * to see various widgets or notices.
+	 *
+	 * @since 1.9.3
+	 */
```
SteveJonesDev, line 220: add @return parameter:

` * @return bool True if the current user has the required capability, false otherwise.`

admin/class-helpers.php
```diff
@@ -211,4 +211,21 @@ function ( $rule ) {
...
+	public static function current_user_can_see_widgets_and_notices(): bool {
```
SteveJonesDev, line 221: Shouldn't we create a unit test for this method?

SteveJonesDev (changes requested):

admin/class-helpers.php
pattonwebz: Good idea, I added tests for expected values and also a case that covers what would be expected should the value be filtered to a given cap

SteveJonesDev (approved):

Issue #535, opened 2024-03-15
Post edit move to trash button causes error and prevents issue purge


SteveJonesDev:
* [x] I've read and understood the [contribution guidelines](https://github.com/equalizedigital/accessibility-checker/blob/develop/.github/CONTRIBUTING.md).
* [x] I've searched for any related issues and avoided creating a duplicate issue.

### Please give us a description of what happened.
The `Move to trash` button on the post-edit screen causes the plugin to trigger a new scan which fails and also fails to run purge methods.

https://github.com/equalizedigital/accessibility-checker/assets/2895788/076e3c4e-7d0d-4ccf-a539-01b61b4caf57

### Please describe what you expected to happen and why.
Clicking the `Move to trash` button shouldn't initial a new scan and should successfully run the purge actions.

### How can we reproduce this behavior?
1. Install EDAC
2. Create a post
3. Click the `Move to trash` button in the post edit screen

### Technical info
* WordPress version: 6.4.3
* Accessibility Checker version: https://github.com/equalizedigital/accessibility-checker/pull/512

* If relevant, which editor is affected (or editors): 
- [ ] Classic Editor
- [x] Gutenberg
- [ ] Classic Editor plugin
- [ ] Page Builder (please specify):

* Which browser is affected (or browsers): 
- [ ] Internet Explorer
- [ ] Edge
- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [x] Brave

* Tested with theme: TwentTwentyFour

pattonwebz:
I could reproduce this on a local test site as well. When posts are trashed from the block editor, they hit the `save_post` action (for some reason, unsure why), triggering our save flow.

It reaches `edac_validate()`, which tries to build a `simple_html_dom` object from the URL - a 404 by that point - resulting in the 'password protected' notice being thrown up.

Looking at the $post object status, by the time it reaches `edac_save_post`, is already set to `trash` so we can just check for that in the function and return early before attempting to scan the no-longer-accessible post. 

Early return prevents the being thrown and doesn't halt the rest of the flow which deletes the rows from the table.

https://github.com/equalizedigital/accessibility-checker/assets/3902039/3c941730-5989-4b6b-9e9b-f8ef4d61c30c

SteveJonesDev:
That check is potentially a good failsafe. We should still identify why `edac_save_post` is being called when that button is clicked.

pattonwebz:
[...]

PR #536, opened 2024-03-15
Short-circuit edac_save_post() when post_status is trash

pattonwebz:
This is a first pass at a solution for #535 . It returns from the save method when the post is trashed to avoid it, resulting in request errors later in the chain.

Bailing early here ensures that the rest of the flow runs and the data is purged.

There may be a better solution to this, but I haven't found anything as simple as this. I will continue to think and find something more robust if I can.

Resolves: #535

### DIFF
--- a/includes/validate.php
+++ b/includes/validate.php
@@ -51,6 +51,8 @@ function edac_post_on_load() {
  * @param object $post    The post object being saved.
  * @param bool   $update  Whether this is an existing post being updated.
  *
+ * @modified 1.10.0 to add a return when post_status is trash.
+ *
  * @return void
  */
 function edac_save_post( $post_ID, $post, $update ) {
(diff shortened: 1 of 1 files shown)


SteveJonesDev (changes requested):
This fixes the error log output, but I still see a momentary error from the API call.

Screenshot 2024-03-18 at 10 39 43 AM

pattonwebz:
> This fixes the error log output, but I still see a momentary error from the API call.
> 
> Screenshot 2024-03-18 at 10 39 43 AM

I did not see this on my test site, but it appeared after installing Yoast SEO and remains even after disabling Accessibility-Checker. Do you have Yoast SEO installed on your test site?

I can reproduce it on a fresh test site with just Yoast SEO installed. I opened an issue on their repo, but I am confident this issue does not come from Accessibility Checker.

Issue #467, opened 2024-01-25
Convert meta boxes into a class


SteveJonesDev:
https://github.com/equalizedigital/accessibility-checker/blob/develop/includes/meta-boxes.php

Should be an `EDAC\Admin` class.

PR #537, opened 2024-03-18
Convert editor meta boxes function into a class

pattonwebz:
* Moves the registration and render of the main editor metabox to a class.
* Starts a dependency injection pattern for passing classes into the main `Admin` class (makes the class and what happens to it more testable)
* Adds 2 actions around the rendering of the metabox incase there is a need to extend what's rendered.

Closes: #467


SteveJonesDev (approved):
Tested and appears to work as expected. This is good to merge.

Issue #539, opened 2024-03-20
Auto-draft on a new post triggers a scan resulting in an error


pattonwebz:
### Please give us a description of what happened.

When you create a new post in the block editor a scan is triggered on the auto draft resulting in an an error that causes the password protected message to appear until a post is saved and a scan on a valid page is successful.

### Please describe what you expected to happen and why.

No scan is triggered on an autodraft with no content.

### How can we reproduce this behavior?
1. Create a new post
2. See message in error logs
3. Notice the password protected site message banner appears till a post is published or it's otherwise cleared.

### Technical info
* WordPress version:
* Accessibility Checker version:

* If relevant, which editor is affected (or editors): 
- [ ] Classic Editor
- [x] Gutenberg
- [ ] Classic Editor plugin
- [ ] Page Builder (please specify):

* Which browser is affected (or browsers): 
- [ ] Internet Explorer
- [ ] Edge
- [x] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Brave

pattonwebz:
There is a function in the plugin that triggers a frontend validation scan on a post. It's triggered by `template_redirect` hook.

The block editor, when opened, makes a call to determine the available templates the post_type can use which triggers that action and the validation scan. If the editor is opened for a new post it will be an auto-draft and have no accessible page yet so the validate fails resulting in a logged error and then the plugin makes the assumption that the site is password protected.

Error that is thrown: 
```
[19-Mar-2024 18:08:33 UTC] PHP Warning:  file_get_contents(http://local.testsite.test/?p=62642&edac_cache=1710871710): Failed to open stream: HTTP request failed! HTTP/1.1 404 Not Found
 in /var/www/html/wp-content/plugins/accessibility-checker/includes/simplehtmldom/simple_html_dom.php on line 79
```

SteveJonesDev:
Related issue: https://github.com/equalizedigital/accessibility-checker/issues/476

PR #540, opened 2024-03-20
Auto draft causing error attempting scan not-existing page

pattonwebz:
Adds a catch inside the Frontend_Validate classes validate method to return before attempting to validate if the post is an auto-draft.

Fixes: #539


SteveJonesDev (approved):
Good to merge.

Issue #544, opened 2024-03-25
gifs or webp with missing url scheme cause error when checking if they are animated


pattonwebz:
### Please give us a description of what happened.
A scan running against a gif or webp image to check if it is animated fails, it is not able to open the file because of a missing protocol. This blocks the saving of content and throws visible errors to the user.

### Please describe what you expected to happen and why.
Plugin handles the missing scheme and does not show error.

### How can we reproduce this behavior?
1. Add `<img src="//assets.rewardstyle.com/images/search/350.gif">` to a post
2. Save the post
3. See error notice and fail to save message.

### Technical info
* WordPress version:
* Accessibility Checker version: 1.10.0

* If relevant, which editor is affected (or editors): 
- [ ] Classic Editor
- [x] Gutenberg
- [x] Classic Editor plugin
- [ ] Page Builder (please specify):

* Which browser is affected (or browsers): 
- [ ] Internet Explorer
- [ ] Edge
- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Brave

* Relevant plugins in case of a bug:

* Tested with theme:

PR #545, opened 2024-03-25
Handle missing scheme in animated image rule

pattonwebz:
The rule that handles detection of animated gifs and webp images was causing an error on sites when the image src was missing a scheme. This PR solves for that by creating a helper function to add a scheme and verify the URL is accessible before opening the file to analyse it's frames.

Fixes: #544


includes/helper-functions.php
```diff
@@ -857,3 +857,83 @@ function edac_database_table_count( $table ) {
...
+	if ( ! str_starts_with( $filename, '//' ) ) {
+		$file = $filename;
+	} else {
+		$file       = edac_url_add_scheme_if_not_existing( $filename );
+		$url_exists = edac_url_exists( $file );
+		// if the https version does not exist, try http.
+		if ( false === $url_exists ) {
+			$file       = edac_url_add_scheme_if_not_existing( $filename, 'http' );
+			$url_exists = edac_url_exists( $file );
+		}
```
pattonwebz, line 923: I thought about this over the weekend and I want to just do away with the nested if logic here and instead assume it will always have only what the site protocal is since the page it loads on would dictate it. Thoughts on dropping the internal if here that falls back to `http`?

SteveJonesDev: Yes, we can remove this logic check.

SteveJonesDev (changes requested):

SteveJonesDev (changes requested):
URL's like this `assets.rewardstyle.com/images/search/350.gif` are still being opened without a protocol.

Here is the warning:

```
[26-Mar-2024 16:39:23 UTC] PHP Warning:  fopen(rewardstyle.com/images/search/350.gif): Failed to open stream: No such file or directory in /Users/stevejones/Documents/Repositories/accessibility-checker/includes/helper-functions.php on line 950
```

SteveJonesDev (approved):
Good to merge.

Issue #548, opened 2024-03-25
Reading density data can sometimes result in fatal error


pattonwebz:
* [x] I've read and understood the [contribution guidelines](https://github.com/equalizedigital/accessibility-checker/blob/develop/.github/CONTRIBUTING.md).
* [x] I've searched for any related issues and avoided creating a duplicate issue.

### Please give us a description of what happened.
Saving a post gave one user a fatal error, and they could reproduce it at least one additional time. The issue was caused by a string reaching a place in the summary data calculations that expected an array but got a string.

### Please describe what you expected to happen and why.
Posts can be saved without causing any errors.

### Technical info
* Accessibility Checker version: 1.10.0

* If relevant, which editor is affected (or editors): 
- [ ] Classic Editor
- [x] Gutenberg
- [x] Classic Editor plugin
- [ ] Page Builder (please specify):

PR #549, opened 2024-03-25
Check we have an array before accessing it when producing density to return via ajax

pattonwebz:
When certain validation errors occur a default density_data meta value is created as a string of `'0,0'` but when the data is processed it expects it to be an array. 

* Updates those places in the validation method to use an array of `[ 0, 0 ]` so it matches what is expected when processing.
* Adds some additional checks before processing or accessing it to ensure it is an array.
* Adds a test case that validates the issue is fixed

Fixes: #548

### DIFF
--- a/includes/validate.php
+++ b/includes/validate.php
@@ -362,12 +362,12 @@ function edac_get_content( $post ) {
 				}
 			}
 		} catch ( Exception $e ) {
-			update_post_meta( $post->ID, '_edac_density_data', '0,0' );
+			update_post_meta( $post->ID, '_edac_density_data', array( 0, 0 ) );
 
 			$content['html'] = false;
 		}
 	} else {
-		update_post_meta( $post->ID, '_edac_density_data', '0,0' );
+		update_post_meta( $post->ID, '_edac_density_data', array( 0, 0 ) );
 
 		$content['html'] = false;
 	}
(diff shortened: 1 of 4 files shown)


includes/validate.php
```diff
@@ -362,12 +362,12 @@ function edac_get_content( $post ) {
 				}
 			}
 		} catch ( Exception $e ) {
-			update_post_meta( $post->ID, '_edac_density_data', '0,0' );
+			update_post_meta( $post->ID, '_edac_density_data', array( 0, 0 ) );
```
SteveJonesDev, line 365: I'm curious why Matt had this set as a string rather than an array and if that was intentional. I did a quick test with the Audit History add-on and I didn't notice any issues. Do you thing it needs further investigation?

SteveJonesDev (changes requested):

includes/validate.php
pattonwebz: I dug back through past versions of the files that set and use this and it seems as though it's been expected to be an array with 2 keys going back a very long way. 

I suspect that this was just added as a catch-all default value and wasn't easy to trigger due to this code only executing under uncertain error conditions that are hard to replicate. This being a string here seems like just an assumption that couldn't be tested under normal operations. It would always have caused issues if it stored this to be read later on.

pattonwebz: This issue is the last time this code was edited but even back then it expected that the value would be an array returned with this array and 2 keys inside it. https://github.com/equalizedigital/accessibility-checker/issues/245

SteveJonesDev (approved):

Issue #522, opened 2024-03-06
False positives on Duplicate Alt Text


amberhinds:
* [x] I've read and understood the [contribution guidelines](https://github.com/equalizedigital/accessibility-checker/blob/develop/.github/CONTRIBUTING.md).
* [x] I've searched for any related issues and avoided creating a duplicate issue.

### Please give us a description of what happened.

I'm seeing lots of false positives on duplicate alt text, including on images that don't have alt text at all. Here are some examples.

This is on our website - a Facebook pixel icon
<img width="1079" alt="Screenshot 2024-03-06 at 5 16 16 PM" src="https://github.com/equalizedigital/accessibility-checker/assets/13153456/946629c1-feec-459f-b9a2-02dc8246a071">

This is on a customer's website - SVG missing alt. (It does also flag Missing Alt errors)
<img width="821" alt="Screenshot 2024-03-06 at 10 36 53 AM" src="https://github.com/equalizedigital/accessibility-checker/assets/13153456/1d8521a7-f325-42e4-a8de-1b3157ec5242">

### Please describe what you expected to happen and why.
No false positives. 😜

### How can we reproduce this behavior?
Check the report on https://equalizedigital.com/wp-admin/post.php?post=4991&action=edit

### Technical info
* WordPress version:
* Accessibility Checker version: 1.9.1
* If relevant, which editor is affected (or editors): 
- [ ] Classic Editor
- [x] Gutenberg
- [ ] Classic Editor plugin
- [x] Page Builder (please specify): Beaver Builder (customer example)

* Tested with theme: Multiple.

pattonwebz:
I have a PR up that resolves the false flagging of duplicate alt text when there was no alt on the images. The problem was expecting an empty string in a strict comparison when the return could be `false` if the attribute didn't exist.

christophermhinds:
@pattonwebz @SteveJonesDev 

🟡 At least in the one test case I'm aware of in the Equalize Digital website footer, it is still flagging that Facebook pixel as having duplicate alternative text after installing the test release and re-scanning the page. See the screenshot below. Looking at it in the front-end highlighting view, it indicates that the element is not visible on the page. I am not sure if that's relevant or not.

<img width="1023" alt="Screenshot 2024-04-30 at 11 17 49 AM" src="https://github.com/equalizedigital/accessibility-checker/assets/25856499/5df49e7e-fb6a-4940-8a8c-3be6c7fcc3a8">

PR #550, opened 2024-03-26
Prevent false flagging duplicate alt test when alt isn't present

pattonwebz:
The check rule for redundant alt text was flagging elements that had no alt attribute at all because it was checking for `!== ''` (not empty strings) when Simple HTML Dom returns `false` for non-existent alt attribute causing it to not return when intended.

* Updates some condition checks to use `! empty()` so it understands `false` like intended - instead of checking for empty string only.

Fixes: #522


includes/rules/img_alt_redundant.php
```diff
@@ -8,6 +8,8 @@
 /**
  * IMG ALT Redundant
  *
+ * @modified 1.11.0 Updated various conditions to use `empty` checks instead of checking for empty strings, since SimpleHTMLDom returns `false` for non-exist` attributes.
```
pattonwebz, line 11:
```suggestion
 * @modified 1.11.0 Updated various conditions to use `empty` checks instead of checking for empty strings since Simple HTML Dom returns `false` for non-existent attributes.
```

tests/phpunit/TestHelpers/GetDomHelperTrait.php
```diff
@@ -0,0 +1,31 @@
...
+trait GetDomHelperTrait {
+
+	/**
+	 * Wrapper to generate dom objects that match the shape of the object in the plugin.
+	 *
+	 * @param string $html_string HTML string.
+	 * @return EDAC_Dom
+	 */
+	public function get_DOM( string $html_string = '' ): EDAC_Dom { // phpcs:ignore WordPress.NamingConventions -- This is a helper function.
+		return new EDAC_Dom(
```
SteveJonesDev, line 21: I don't think this rule used the `EDAC_Dom` class. Is there a reason to use this over using `str_get_html()`?

SteveJonesDev (changes requested):
Tested and it appears to fixes the false positives. Everything looks good, I just have a question about the use of the `EDAC_Dom` class in the tests.

tests/phpunit/TestHelpers/GetDomHelperTrait.php
pattonwebz: The `str_get_html()` likely has the same string generated as what this is for. I used this though because I expected that it would be best to get the html string in exactly the same way as it would be passed to the rule function from in the plugin.

One way in the plugin, one way in the tests and they are the same is the general idea. It wouldn't make a huge amount of difference if I dropped this in favour of just calling `str_get_html()` though and I can do that going forward if you feel strongly about simplifying in that way.

PR #552, opened 2024-03-26
Release/v1.10.1

pattonwebz:
Steve Jones (1):
- [x] Missing form label rule strict comparison fix

William Patton (6):
- [x] Convert editor meta boxes function into a class
- [x] Short-circuit edac_save_post() when post_status is trash
- [x] Fix missing clarifying parenthesis cs check
- [x] Check we have an array before accessing it when producing density to return via ajax
- [x] Handle missing scheme in animated image rule


readme.txt
SteveJonesDev, line 1: The stable tag needs to be updated to: 1.10.1

SteveJonesDev (changes requested):
Just need to update the stable tag.

SteveJonesDev (approved):

Issue #402, opened 2023-12-08
Use fewer postmeta values


jdevalk:
This:

```php
	// save summary data as post meta.
	if ( ! add_post_meta( $post_id, '_edac_summary', $summary, true ) ) {
		update_post_meta( $post_id, '_edac_summary', $summary );
	}

	if ( ! add_post_meta( $post_id, '_edac_summary_passed_tests', $summary['passed_tests'], true ) ) {
		update_post_meta( $post_id, '_edac_summary_passed_tests', $summary['passed_tests'] );
	}

	if ( ! add_post_meta( $post_id, '_edac_summary_errors', $summary['errors'], true ) ) {
		update_post_meta( $post_id, '_edac_summary_errors', $summary['errors'] );
	}

	if ( ! add_post_meta( $post_id, '_edac_summary_warnings', $summary['warnings'], true ) ) {
		update_post_meta( $post_id, '_edac_summary_warnings', $summary['warnings'] );
	}

	if ( ! add_post_meta( $post_id, '_edac_summary_ignored', $summary['ignored'], true ) ) {
		update_post_meta( $post_id, '_edac_summary_ignored', $summary['ignored'] );
	}

	if ( ! add_post_meta( $post_id, '_edac_summary_contrast_errors', $summary['contrast_errors'], true ) ) {
		update_post_meta( $post_id, '_edac_summary_contrast_errors', $summary['contrast_errors'] );
	}
```

could be one post meta value containing an array instead of 6, saving loads and loads of post meta table rows.

SteveJonesDev:
The reason these are saved as individual post meta is for the custom column sorting in the Pro plugin. https://github.com/equalizedigital/accessibility-checker-pro/blob/develop/includes/custom-columns.php#L71-L116

 The post meta `_edac_summary` is a serialized array in the database, and WordPress doesn't support ordering by serialized array values out of the box. We'll need to look at maybe writing a custom MySQL function or some other solution.

aristath:
Actually, we can get rid of the individual post-meta, and then implement custom ordering logic without having to write a custom SQL query or function (something which admittedly would overcomplicate things).
Instead, we can manipulate the `$wp_query` global var in the page, and reorder the posts in it.

[...]

PR #578, opened 2024-04-08
Use Fewer post meta values

pattonwebz:
This PR aims to reduce the number of different post meta values in favor of storing as much data inside of a single post meta as possible.

The exception is that for sorting by columns some meta still must exist in 2 places.

It reduces the total number of post meta by 5 and introduces a single interface to getting any piece of post meta.

https://github.com/equalizedigital/accessibility-checker/pull/578/files#diff-ac3197d4f34d0b1b8fafdf64365bd3e7cb32db3aa83571ea74fcdfe464adf0eeR118-R122

Closes: #402

### DIFF
--- a/admin/class-ajax.php
+++ b/admin/class-ajax.php
@@ -7,7 +7,7 @@
 
 namespace EDAC\Admin;
 
-use EDAC\Admin\Data\Post_Meta\Scan_Summary;
+use EDAC\Admin\Data\Post_Meta\{ Scan_Summary, Scan_Summary_Back_Compat };
 use EDAC\Inc\Summary_Generator;
 
 /**
@@ -77,7 +77,7 @@ class Ajax {
 		$post_id                   = (int) $_REQUEST['post_id'];
 		$summary                   = ( new Summary_Generator( $post_id ) )->generate_summary();
 		$simplified_summary_prompt = get_option( 'edac_simplified_summary_prompt' );
-		$simplified_summary        = ( new Scan_Summary( $post_id ) )->get( 'simplified_summary_text' );
+		$simplified_summary        = ( new Scan_Summary_Back_Compat( $post_id ) )->get( 'simplified_summary_text' );
 
 		$simplified_summary_grade = 0;
 		if ( class_exists( 'DaveChild\TextStatistics\TextStatistics' ) ) {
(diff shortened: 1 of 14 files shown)


SteveJonesDev:
@pattonwebz, here is more context to the full site scanning not working that @christophermhinds mentioned.

Free: 1.10.2
Pro: 1.6.0
Full site scan works as expected

Free: 1.11.0-alpha.1
Pro: 1.6.0
Full site scan pass two doesn't run
No console or log errors

Free: 1.11.0-alpha.1
Pro: Version 1.7.0-alpha.1
Full site scan pass two doesn't run
No console errors
It does log a deprecation. Should it be doing if both versions are the current alpha? Maybe you need to check for the edac_summary for backward compatibility?
`[18-Apr-2024 18:02:19 UTC] PHP Deprecated:  Function edac_summary is deprecated since version 1.9.0! Use EDAC\Inc\Summary_Generator instead. in /Users/stevejones/Local Sites/edactest/app/public/wp-includes/functions.php on line 6078`

Free: 1.10.2
Pro: 1.7.0-alpha.1
Full site scan works as expected
Same deprecation log.
`[18-Apr-2024 18:02:19 UTC] PHP Deprecated:  Function edac_summary is deprecated since version 1.9.0! Use EDAC\Inc\Summary_Generator instead. in /Users/stevejones/Local Sites/edactest/app/public/wp-includes/functions.php on line 6078`

These findings lead me to believe the issue is in Free: 1.11.0-alpha.1

SteveJonesDev (changes requested):
`_edac_issue_density` and `_edac_post_checked_js` aren't being migrated into `_edac_summary`. Maybe that's for a reason?

Also, `_edac_issue_density` does exist in `_edac_summary` as `issue_density` but the values don't match.

Before:
Screenshot 2024-04-18 at 2 49 57 PM

After:
Screenshot 2024-04-18 at 2 50 53 PM

Full site scan not working detailed here: https://github.com/equalizedigital/accessibility-checker/pull/578#issuecomment-2064813327

Add a page to the Wiki detailing the proper usage of meta in the plugin.

pattonwebz:
@SteveJonesDev I found the issue with the JS scan not working but it's proving difficult to find a solution that doesn't involve a rebuild of much of the JS scan checking logic in pro. It uses some direct queries against meta keys to count posts and more direct queries to update the values.

It probably means that the `_edac_post_checked_js` meta is another key that will need to live outside of the single meta - unless a lot of attention is given to the JS scan logic in pro to refactor avoiding the need for the dedicated meta key. I know that refactoring of the pro plugin is coming up, so I think I would propose refactoring that logic then as part of the overall plugin refactoring work. I can make a card on the pro repo.

pattonwebz:
After some discussion and undertaking the work required to reduce post meta, it was decided to postpone this. The reduction in post meta rows was minor, but the changes were far-reaching (14 files changed in this PR and the value gained is small. These changes don't entirely account for all the post meta used by the pro plugin, which is an even more significant chunk.

After reviewing the work, it is clear that to get the most value from the changes, it requires redesigning how the suite of plugins handles data entirely, particularly column filtering and how the full site scan—the JS portion especially—deals with their status and tracking. This would be a breaking change requiring a major version bump and not something that can be squeezed into a typical release.

pattonwebz:
Closing this PR now, it is diverged too much to be useful and in the end I don't think we can tackle this in this way anymore. It would need a ground up rethink of data handling rather than a top down refactor.

Issue #580, opened 2024-04-09
Can creation of custom nonce be avoided?


pattonwebz:
There is one place in the validation flow where a shortlived nonce is created via custom methods. We should try to use WordPress nonce creation and validation instead of reimplementing this in the plugin. 

https://github.com/equalizedigital/accessibility-checker/blob/develop/includes/helper-functions.php#L458-L503

SteveJonesDev:
Issue moved to our internal system.

PR #581, opened 2024-04-10
Use wp nonce methods rather than custom ones in the validate flow

pattonwebz:
The validation flow used a custom-generated short-lived nonce to get the HTML content of draft or pending pages.

An actual WP nonce can be used instead here, which is what this PR swaps over to use.

The custom nonce was needed because the request didn't have any user context, but we can give it the user context so the wp nonce validates by forwarding the cookies in the request. Since this flow can only happen through a logged-in user that seems reasonable to me.

Closes: #580

### DIFF
--- a/includes/validate.php
+++ b/includes/validate.php
@@ -331,6 +331,9 @@ function edac_get_content( $post ) {
 			// Add the token to the URL.
 			$url = add_query_arg( 'edac_token', $token, $url );
 
+			// Since this uses a wp nonce we need to send along the cookies of the user making the request.
+			$context_opts['http']['header'] = 'Cookie: ' . edac_get_current_user_cookies();
+
 		}
 
 		try {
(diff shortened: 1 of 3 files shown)


includes/validate.php
```diff
@@ -465,10 +470,40 @@ function edac_show_draft_posts( $query ) {
...
+ * Get current user cookies
+ *
+ * @return string
+ */
+function edac_get_current_user_cookies(): string {
+	$cookies = array();
+	// phpcs:ignore WordPressVIPMinimum.Variables.RestrictedVariables.cache_constraints___COOKIE -- User must be logged in to trigger this so it's a valid use for cookies on the server.
+	foreach ( $_COOKIE as $key => $value ) {
+		$cookies[] = $key . '=' . $value;
+	}
```
pattonwebz, line 491: This might be why not doing this before was considered —VIP rules flag that cookies may not be reliable due to caching. I don't think that's a concern, though; the user can trigger this from a logged-in state only, where the session cookie must exist and be transferred on requests.

SteveJonesDev:
We ran into some complex issues with cookies not always passing properly maybe due to caching in some environments. Also, previously these scans could run on a schedule in the background adding to the complexity of passing user cookies. I'm going to table this for now and we'll revisit it.

pattonwebz:
We won't be moving forward with this anytime soon so I am closing this out but we can revisit it in future.

PR #583, opened 2024-04-16
Show email opt-in as modal for users first visit to welcome page

pattonwebz:
This PR introduces an opt-in modal that shows on the welcome screen the first time any user visits it - providing they are not already flagged as signed up via meta created on form submit.

It does not handle redirecting users to the welcome page on activation as a recent core change made that no longer possible using the plugin activation hook. See: https://core.trac.wordpress.org/ticket/60992 We should wait for a resolution (or alternative methods for it) to that ticket before trying to add a redirect.

* Adds a class to handle bulk of the email operations.
* Moves markup and checking logic out  of Welcome_Page to new email class


SteveJonesDev (changes requested):
Overall the code looks good.

The 'edac_email_optin_seen_modal` should save as a single value. Sometimes it saves an array of values.
Screenshot 2024-04-17 at 3 36 50 PM

The 'edac_email_optin_seen_modal' and `edac_email_optin` user meta should be purged on uninstall if the `edac_delete_data` option is checked in the settings.
See: https://github.com/equalizedigital/accessibility-checker/blob/develop/uninstall.php#L13-L44

pattonwebz:
> The `edac_email_optin_seen_modal` and `edac_email_optin` user meta should be purged on uninstall if the `edac_delete_data` option is checked in the settings.

I can add this but I will need to do it via a DB Query so I can do it in one query - and not by iterating through users. On a site with a lot of users iterating through them all and doing hundreds of database queries wouldn't be good.

We also aren't deleting post meta either, which probably should have an issue created for it, as I don't think I want to do that in this PR.

SteveJonesDev:
You make a good point. Let's push the deleting of the user and post meta to separate issues and come up with a good solution for that. As you said this could be a very taxing process on sites with a large number of users and posts.

Issue #577, opened 2024-04-07
False Positives on Underline Text - not accounting for :not


amberhinds:
### Please give us a description of what happened.

I received a report that the Accessibility Checker incorrectly flagged underlined text when text-decoration was applied to links. I believe the problem is that we're not accounting for the use of [:not()](https://developer.mozilla.org/en-US/docs/Web/CSS/:not). This is the specific CSS being flagged:

`#main-content a:not(.et_pb_button):not(.button):not(.btn):not(.et-pb-arrow-prev):not(.et-pb-arrow-next){ text-decoration: underline; }`

As you can see this is applying underline to links and this should not create an issue.

### How can we reproduce this behavior?

This is the customer's website: https://fxbg.com/our-hearts-sparkle-for-these-veterans-in-fredericksburg-throughout-february/

Go to https://demosite.equalizedigital.com/wp-admin/post.php?post=780&action=edit - I added the content and CSS there so you can see the warning getting flagged.

amberhinds:
I debated between a medium and low priority for this, but ultimately went for medium because I know of several users who have reported false positives on this check, including the VA. Feel free to adjust if you think it should be different @pattonwebz.

Issue #344, opened 2023-10-20
Underlined Text False Positive


amberhinds:
<img width="1096" alt="Screenshot 2023-10-19 at 10 20 56 PM" src="https://github.com/equalizedigital/accessibility-checker/assets/13153456/52a5e76b-6009-4f16-8e47-b70bf1280776">

These look like false positives because they are targeting links.

christophermhinds:
@pattonwebz @SteveJonesDev 

✅ I tested the underline text rule in a couple of different contexts and it appears to correctly flag underlined text with a warning.

Issue #266, opened 2023-09-05
Incorrectly flagged underlined text


boonedev:
The underlined_text rule can incorrectly parse css and falsely flag content even though it is not underlined. See this page for an example: https://bricks4wp.com/general/replacing-rss-with-wp-api/. 

To prevent the css parsing errors, the rule needs to be ported from php to js.

christophermhinds:
See [11/20/23 Team Meeting Notes - Internal](https://3.basecamp.com/3579237/buckets/1722349/messages/6773425598) for more information / related discussion on this issue.

christophermhinds:
@pattonwebz @SteveJonesDev 

✅ I tested the underline text rule in a couple of different contexts and it appears to correctly flag underlined text with a warning.

PR #586, opened 2024-04-23
Swap underlined_text rule to js scan so that we can get more accurate computed styles

pattonwebz:
Adds a custom axe-core rule and checks to cover the detection from the php check for the same thing.

 Moving to axe-core and JS checking allows more accurate detection of underlines because we can do computed style checks on the actually rendered content - no longer need to try and parse then figure out CSS rules and how they are applied. We get the exact styles computed for the specific element.

Fixes: #577
Fixes: #344 
Fixes: #266


pattonwebz:
In the interest of prepping the beta I am going to merge this without waiting on code review. This was already loosely tested but I would still appreciate code review when you have the chance.

SteveJonesDev (commented):
Looks good.

PR #591, opened 2024-04-26
Add a rule and check for empty paragraphs

pattonwebz:
This introduces a new rule to the JS scanner that can detect empty paragraph tags (or tags with whitespace inside of them, like spaces or returns).

@amberhinds I am going to need some copy/docs for this scan since it is new and there isn't something existing I can use or point to. An info URL and improved summary to show users in the details view and frontend highlighter. Picking which tags are suitable as well will be needed: https://www.deque.com/axe/core-documentation/api-documentation/#axecore-tags


amberhinds:
When do you think this rule will be released @pattonwebz?

pattonwebz:
@amberhinds it can be in the next release (next Tuesday) but if that's too soon we can probably do the week after? I do have one other new rule I will be adding as well that would need docs too.

includes/rules.php
```diff
@@ -74,6 +74,14 @@
 			'<code>&lt;h2&gt;</code>'
 		),
 	),
+	array(
+		'title'     => esc_html__( 'Empty Paragraph Tags', 'accessibility-checker' ),
```
amberhinds, line 78: @pattonwebz please make title singular - "Empty Paragraph Tag" - this will match how we have named other rules.

includes/rules.php
```diff
@@ -74,6 +74,14 @@
...
+		'info_url'  => 'https://a11ychecker.com/help###',
+		'slug'      => 'paragraph_should_not_be_empty',
+		'rule_type' => 'error',
+		'summary'   => esc_html__( 'Paragraph tags that are empty get announced to screen readers with empty content creating unnecessary noise. Paragraphs should never be used for layout purposes.', 'accessibility-checker' ),
```
amberhinds, line 82: An Empty Paragraph Tag error means there is a paragraph tag present that does not contain content. These are announced by screen readers and create unnecessary noise for blind users. To fix this error, remove the empty paragraphs from the page. If you need to add spacing between sections, this should be done with padding, margins, or a spacer block.

amberhinds (changes requested):
@pattonwebz some wording changes.

amberhinds:
Tag: best-practice

This is going to be the documentation link: https://equalizedigital.com/accessibility-checker/empty-paragraph-tag/

I will finish that up in the morning. Let me know if you need anything esle.

includes/rules.php
```diff
@@ -74,6 +74,14 @@
 			'<code>&lt;h2&gt;</code>'
 		),
 	),
+	array(
+		'title'     => esc_html__( 'Empty Paragraph Tag', 'accessibility-checker' ),
+		'info_url'  => 'https://a11ychecker.com/',
+		'slug'      => 'paragraph_should_not_be_empty',
```
SteveJonesDev, line 80: Can we change this to `empty_paragraph_tag ` to be consistent with other rule slugs?

src/pageScanner/index.js
```diff
@@ -3,6 +3,9 @@
 
 import 'axe-core';
 import colorContrastFailure from './rules/color-contrast-failure';
+import paragraphShouldNotBeEmpty from './rules/paragraph-should-not-be-empty';
```
SteveJonesDev, line 6: Can we change this to `empty_paragraph_tag`  as well?

SteveJonesDev (changes requested):

includes/classes/class-rest-api.php
```diff
@@ -161,7 +162,27 @@ function () use ( $ns, $version ) {
 		);
 	}
 
-
+	/**
+	 * Filter the html of the js validation violation.
+	 *
+	 * This can be used to store additional data in the html of the violation.
+	 *
```
pattonwebz, line 169:
```suggestion
	 * This can be used to store additional data in the html of the violation.
	 *
	 * @since 1.13.0
```

includes/rules.php
```diff
@@ -74,6 +74,14 @@
 			'<code>&lt;h2&gt;</code>'
 		),
 	],
+	[
+		'title'     => esc_html__( 'Empty Paragraph Tag', 'accessibility-checker' ),
+		'info_url'  => 'https://equalizedigital.com/accessibility-checker/empty-paragraph-tag/',
```
pattonwebz, line 79:
```suggestion
		'info_url'  => 'https://a11ychecker.com/help7870',
```

PR #592, opened 2024-04-26
Add error for zoom disabled via viewport meta tag

pattonwebz:
This will flag an error if user-scalable is set to 'no' or if the maximum-scale is lower than 2. It uses the axe-core `meta-viewport` rule and requires no special configuration. It already has tags so we will only need to create docs and provide a summary message before we release this.

If you want to test this on a block theme like TwentyTwentyThree you will need to use an action to modify the viewport tag. You can use this one (and yes you do need a negative priority and to remove the action on wp_head inside of wp_head).

```
add_action(
	'wp_head',
	function () {
		remove_action( 'wp_head', '_block_template_viewport_meta_tag', 0 );
		echo '<meta name="viewport" content="width=device-width, initial-scale=1, user-scalable=no">';
	},
	PHP_INT_MIN
);
```


SteveJonesDev (approved):

includes/rules.php
```diff
@@ -420,4 +420,12 @@
 			'<code>&lt;a&gt;</code>'
 		),
 	],
+	[
+		'title'     => esc_html__( 'Zooming and scaling is disabled', 'accessibility-checker' ),
+		'info_url'  => 'https://a11ychecker.com/help###',
```
SteveJonesDev, line 425: Pending documentation.

amberhinds:
@SteveJonesDev @pattonwebz can we remove the "View on Page" action for this rule? It's never going to work so it seems like we should not offer it.
Screenshot 2024-05-23 at 4 44 51 PM

SteveJonesDev:
Yes, we can do that @amberhinds. @pattonwebz we'll need to do the same in the Pro plugin.

PR #596, opened 2024-04-30
Release/1.11.0

pattonwebz:
- [x] Show email opt-in as modal for users first visit to welcome page
- [x] Don't build wp-env when doing npm install in CI as those containers are not needed
- [x] Point the TextStatistics package to the forked repo
- [x] Use short array syntax
- [x] Update the date of the GAAD discount notice to 2024 year dates
- [x] Swap underlined_text rule to js scan so that we can get more accurate computed styles

This release has a huge number of files changed due to the array swap - so I know it will be hard to review. I think you can just review the underlined_text rule change - make sure it continues to work on a test site.


SteveJonesDev (approved):
Tested and reviewed. It looks ready for release.

Issue #611, opened 2024-05-06
Remove Playground_Check class


SteveJonesDev:
Remove the playground blocking code to test compatibility on the plugin directory.

PR #614, opened 2024-05-06
Remove wp org playground checks

pattonwebz:
Removes the conditional and class used to check for playground and display notice of incompatibility.

Fixes: #611


SteveJonesDev (approved):
Good to merge.

Issue #612, opened 2024-05-06
Possible heading needs to support CSS variables 


amberhinds:
Possible heading doesn't flag on paragraph tags where the font is made larger in core typography styles. See https://a11ycheckrules.wpenginepowered.com/possible-heading/

We should add support for this as a priority.

SteveJonesDev:
If this is a high priority then all rules that require CSS evaluation need to have the same priority for converting them to JavaScript.

amberhinds:
How common do we think CSS variables are? I think pretty common, so maybe all those rules should be converted as a high priority. Also, it means better support for core themes, which seems pretty important.

I think our rules providing highly accurate results is probably the #1 priority because it's one of those baseline, expected things. If the plugin misses stuff compared to other tools, then savvy users won't use it.

PR #622, opened 2024-05-14
Check possible_heading rule with computedStyles

pattonwebz:
This PR moves the possible_heading rule to a JS rule that can check element styling with computedStyles. That prevents any need to parse CSS vars, functions or pseudo-classes since it gets the actual styles the element has after all the cascade is applied.

@amberhinds I will need some help setting the tags that the rule should have.

Sidenote: I like to see PRs that make things more powerful yet are red than green (or orange and blue in my case haha) :) 
Screenshot from 2024-05-14 22-18-28

Fixes: #612


amberhinds:
@pattonwebz what do you mean by tags?

pattonwebz:
@amberhinds sorry I should have linked to the list. I need to know the correct tags and a category from the list here: https://www.deque.com/axe/core-documentation/api-documentation/#axecore-tags

I presume this is the list but let me know if it misses anything or something shuoldn't be in the list:

wcag2a
wcag2aa
wcag21a
wcag21aa
wcag22aa
wcag131  
wcag241
wcag246

amberhinds:
Got it. Here are the tags:

- wcag2a
- wcag131
- wcag241

amberhinds:
@pattonwebz did you use this? https://dequeuniversity.com/rules/axe/4.9/p-as-heading?application=RuleDescription

pattonwebz:
> @pattonwebz did you use this? [dequeuniversity.com/rules/axe/4.9/p-as-heading?application=RuleDescription](https://dequeuniversity.com/rules/axe/4.9/p-as-heading?application=RuleDescription)

I didn't use that rule because it doesn't work as well as what we did in PHP. I replicated our own rule instead of using this. 

The rule from axe-core doesn't check as many things as we do (it doesn't care about font size), and in my testing, the things it found were marked as `incomplete`, which means it needs additional review to determine a pass or fail. It can't tell anything with certainty from the data I have run it against — probably why it's still flagged as an experimental rule.

Screenshot from 2024-05-15 14-56-09

Our rule catches 8 items, and the axe-core rule catches 0 and returns a 'can't tell' result with only one item from my test data.

amberhinds:
Thanks for that info! Sounds like our rule is better. I wonder if we need to document anywhere when our rule is better than an axe rule so we remember that for the future.

pattonwebz:
I can add some comments in the code for this rule so that we can remember the next time we touch the code. I'll see if I can find a place to document it outside of the code for ourselves to and form the basis of some copy for the plugin docs that explains why our check does certain things that axe-core misses.

amberhinds:
Code comments sound good. Outside the code tracking here makes sense: https://docs.google.com/spreadsheets/d/10vbbz4Y4uit76z73U8ffa9h6NN_ha5TC6Vfkt2-4IxI/edit?usp=sharing

src/pageScanner/checks/paragraph-styled-as-header.js
```diff
@@ -0,0 +1,52 @@
...
+ * @return {boolean} True if the node is styled like a header, false otherwise. Paragraphs with only bold or italic,
+ *                   are shorter than 50 characters, or are short with large font size are considered headers.
+ */
+
+import { fontSizeInPx } from '../helpers/helpers.js';
+
+export default {
+	id: 'paragraph_styled_as_header',
+	evaluate: ( node ) => {
+		const pixelSize = fontSizeInPx( node );
```
SteveJonesDev, line 23: I'm not sure what happens before this but is there any value in returning early here if node is empty?

```js
if ( ! node ) {
     return false;
}
```

src/pageScanner/checks/paragraph-styled-as-header.js
```diff
@@ -0,0 +1,52 @@
...
+
+		// long paragraphs or with size under 16px are unlikely to be headers.
```
SteveJonesDev, line 25:
```suggestion
		// Long paragraphs or those with font size under 16px are unlikely to be headers.
```

src/pageScanner/checks/paragraph-styled-as-header.js
```diff
@@ -0,0 +1,52 @@
...
+		if ( node.textContent.trim().length > 50 || pixelSize < 16 ) {
+			return false;
+		}
+
+		// paragraphs that are 20px or more are probably headers.
```
SteveJonesDev, line 30:
```suggestion
		// Paragraphs that are 20px or more are probably headers.
```

src/pageScanner/checks/paragraph-styled-as-header.js
```diff
@@ -0,0 +1,52 @@
...
+		const fontStyle = style.getPropertyValue( 'font-style' );
+		const isItalic = [ 'italic', 'oblique' ].includes( fontStyle );
+
+		const hasBoldOrItalicTag = node.querySelector( 'b, strong, i, em' ) !== null;
+
+		if ( isBold || isItalic || hasBoldOrItalicTag ) {
+			return true;
+		}
+
+		// didn't find anything indicating this is a possible header.
```
SteveJonesDev, line 49:
```suggestion
		// Didn't find anything indicating this is a possible header.
```

src/pageScanner/rules/possible-heading.js
```diff
@@ -0,0 +1,15 @@
+export default {
+	id: 'possible_heading',
+	selector: 'p',
+	excludeHidden: false,
+	tags: [
+		'cat.text',
```
SteveJonesDev, line 6: Does this still need the actual tags?

SteveJonesDev (changes requested):

Issue #630, opened 2024-05-21
Convert the text_small php rule into a JS checked rule


pattonwebz:
The current text_small rule uses php and parses stylesheets and inline styles to determine the size of the element. The PHP check trips up on the use of CSS vars making it less reliable, and also having to partially guess the cascade. Swapping to a JS based rule allows use of computed styles which are much more reliable as they give the actual value rendered.

PR #633, opened 2024-05-22
Convert the text_small rule to JS check

pattonwebz:
This PR converts the `text_small` rule from PHP to the js checker.

The PHP rule relied on complex CSS parsing, which could trip up or produce a false positive when CSS variables were used, the cascade order was unclear, or non-px units were in use.

The JS check uses actual computed styles to determine the exact rendered size of the text.

Closes: #630


src/pageScanner/rules/text-small.js
```diff
@@ -0,0 +1,22 @@
...
+
+export default {
+	id: 'text_small',
+	impact: 'moderate',
+	selector: 'p, span, small, strong, b, i, h1, h2, h3, h4, h5, h6, a, label, button, th, td, li, div, blockquote, address, cite, code, pre, q, s, sub, sup, u, var, abbr, acronym, del, dfn, em, ins, kbd, samp, time, mark, meter, progress, ruby, rt, rp, wbr, bdi, bdo, br, hr, img, input, select, textarea, output, video, audio, canvas, map, area, object, embed, param, source, track, iframe, table, caption, col, colgroup, thead, tbody, tfoot, tr, th, td, dl, dt, dd, ol, ul, menu, dir, li, figure, figcaption, main, header, footer, nav, section, article, aside, details, dialog, summary, data, time, title, style, script, noscript, template',
+	matches: ( element ) => {
+		// only run checks on elements with text content
+		return element.textContent.trim().length;
+	},
+	tags: [ 'cat.text' ],
```
SteveJonesDev, line 15: Do we need tags for this?

SteveJonesDev (changes requested):

src/pageScanner/rules/text-small.js
pattonwebz: We did, I added some from the docs page

Issue #634, opened 2024-05-22
Convert text_justified rule to JS check


pattonwebz:
Convert the text_justified rule to JS checker to avoid need to parse CSS data through PHP without rendering.

PR #635, opened 2024-05-22
Convert text_justified rule to a JS rule

pattonwebz:
This PR shifts the text_justified php rule over to a JS equivalent so that it can determine rendered/computed styles to check against.

This is currently set up to align with the description of the PHP rule but not necessarily the code of the PHP rule.

* Only checks against `<p>`, `<div>` and `<td>` elements (as per the user-facing description of the rule, the actual PHP code was checking all elements, though).
* Checks only elements over 500 characters in length.

@amberhinds @SteveJonesDev I have two questions.
* Should the rule actually only check those 3 elements or do we want it to be wider? If so the description of the rule will need a tweak.
* Can we reduce the threshold of the check from 500 to something more like 200 so that it checks when the content exceeds around 2 lines of text rather than the 5 or 6 it currently checks at?

Closes: #634


amberhinds:
> as per the user-facing description of the rule, the actual PHP code was checking all elements, though

This was my mistake. We should check everything - I can edit docs.

I agree with the recommendation to reduce the character count to 200.

SteveJonesDev:
@amberhinds we'll open an issue to update the docs.

SteveJonesDev (approved):
Good to merge.

SteveJonesDev:
@amberhinds, I updated the documentation for this rule: https://equalizedigital.com/accessibility-checker/justified-text/

@pattonwebz before merging can you update the summary?
```php
'summary'   => esc_html__( 'A Text Justified warning appears in Accessibility Checker when text with more than 200 characters on your post or page is styled with justified alignment (text-align:justify). To fix this warning, remove the justified styling from the specified text elements.', 'accessibility-checker' ),

```

amberhinds:
Thanks @SteveJonesDev!

PR #637, opened 2024-05-22
Enable the blinks and marquee axe-core rules to power the text_blinking_scrolling rule

pattonwebz:
This PR enables the 'blink' and 'marquee' rules from axe-core and combines the results of those checks under the `text_blinking_scrolling` reporting rule.

Using axe-core rules for blink means no longer parsing the CSS and checking for `text-decoration: blink;`. Any JS side check for it will be unreliable because most common browsers strip it out as it was deprecated long ago and is not supported by the major browsers at this point.


SteveJonesDev (approved):
Good to merge. Great work on the logic added to combine rules.

PR #650, opened 2024-05-30
Updated link_blank rule allowed phrases

### LINKED ISSUE #629: Consider making the check for links that open in new window more generic
A user made [a feature request on the dot org forums](https://wordpress.org/support/topic/warning-about-links-opening-in-new-windows/) about making the new window warning less rigid so that their links (which do indicate it opens on a new tab) to satisfy the warning.

## Is your feature request related to a problem? Please describe.

A link on a page has a label saying it will be a new tab - but the warning is still flagged for it,

## Describe the solution you'd like

A wider check that allows something like `new [tab|window]` to satisfy the issue.

## Why do you think this feature is something we should consider for this plugin?

Not mentioned in the request but it could reduce noise and make the warning more robust against the variance in labeling.

## Additional context

> I have a suggestion/request for the accessibility checker:
>
>The title/aria-label in links which open in a new window which gets flagged is too specific. For example, I have a link on my page that looks like this:
>
>`<a href="..." target="_blank" aria-label="Visit us on Twitter in a new tab">`
>Suggestion: change the text it looks for from “opens in new tab/window” to simply “new tab/window”.
>
>Thanks!

SteveJonesDev:
The PR simplifies the allowed phases used in the Link Opens New Window or Tab rule check.

Closes #629

### DIFF
--- a/includes/rules/link_blank.php
+++ b/includes/rules/link_blank.php
@@ -91,15 +91,11 @@ function edac_check_link_blank_text( $text ) {
 
 	$text = strtolower( $text );
 
-	// phrases.
 	$allowed_phrases = [
-		__( 'opens a new window', 'accessibility-checker' ),
-		__( 'opens a new tab', 'accessibility-checker' ),
-		__( 'opens new window', 'accessibility-checker' ),
-		__( 'opens new tab', 'accessibility-checker' ),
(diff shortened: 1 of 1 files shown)


pattonwebz (approved):
LGTM

Issue #652, opened 2024-06-04
Links modified by the New Window Warning plugin wrongly flagged as possible headers


pattonwebz:
### Please give us a description of what happened.

Reported on dot org forums: https://wordpress.org/support/topic/possible-heading-error-caused-by-new-window-plugin/

Any link modified by the new window warning plugin gets flagged as a possible heading due to them containing `<i>` tags.

### Please describe what you expected to happen and why.

New window links are not flagged as possible headings.

### How can we reproduce this behaviour?
1. Install New Window Warning plugin.
2. Create link that is external.
3. Save the post so it is scanned.
4. See the links flagged as possible headers.

### Technical info
* Accessibility Checker version: 1.13.0

PR #653, opened 2024-06-04
Only consider possible heading if the tag wraps the entire text content

pattonwebz:
Links that are modified by the New Window Warning plugin contain an `<i>` tag, which highlighted the possible header check was checking the containing tags incorrectly. This PR updates it to check if the tag wraps the entire text instead of simply checking it exists inside the scanned node.

Fixes: #652


SteveJonesDev (approved):
Looks good.

Issue #655, opened 2024-06-05
Allow JS rules to keep their ignored state between post updates


pattonwebz:
### Please give us a description of what happened.

A user reported this issue on FB and then posted it on the dot org forums. https://wordpress.org/support/topic/ignored-error-returns-after-page-update/

The error is no longer ignored when a JS-based rule is ignored and the post is saved.

### Please describe what you expected to happen and why.

Errors that are ignored stay ignored, even through post updates.

### How can we reproduce this behaviour?
1. Create an issue in a post that causes a JS rule to flag - like a contrast issue or a possible heading.
2. Save the post and then ignore the contrast issue.
3. Save the post again and see that the issue is no longer ignored.

### Technical info
* Accessibility Checker version: 1.13.0

PR #656, opened 2024-06-05
Don't operate on JS rules when doing rulecheck for php ruleset

pattonwebz:
This PR makes a change that excludes the JS rules from the PHP rule check flow when a post is updated. It prevents any JS rule issues from being deleted from the database at the end of the PHP scan (they will still be deleted if corrected when the JS scan runs).

This solves the issue reported that specific ignored issues (investigating, I found that it was all JS rules subject to this bug) would return when posts were updated and no longer be ignored.

Fixes: #655


SteveJonesDev (approved):
Thanks! This does make it a bit more readable.

PR #662, opened 2024-06-06
Allow frontend highlighter position to be moved from right to left of screen

pattonwebz:
This PR adds a setting that allows people to set the side that the frontend highlighter appears when viewing pages.

It only offers right or left as options. The 3rd option requested (in the admin bar) would require a bigger refactor of the logic for the UI and didn't seem feasible to do quickly.

There are future plans to move the widget to be a sidebar element instead so this change is just a stopgap to prevent our widget interfering with other widget bubbles in the lower right corner on some sites.

### DIFF
--- a/includes/options-page.php
+++ b/includes/options-page.php
@@ -111,6 +111,13 @@ function edac_register_setting() {
 		'edac_settings'
 	);
 
+	add_settings_section(
+		'edac_frontend_highlighter',
+		__( 'Frontend Accessibility Checker', 'accessibility-checker' ),
+		'edac_frontend_highlighter_section_cb',
+		'edac_settings'
+	);
+
 	// Add fields.
 	add_settings_field(
 		'edac_post_types',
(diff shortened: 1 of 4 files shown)


includes/options-page.php
```diff
@@ -285,6 +315,30 @@ function edac_simplified_summary_position_cb() {
...
+ * Renders radio inputs for the frontend highlighter position option.
+ *
+ * @return void
+ */
+function edac_frontend_highlighter_position_cb() {
+	$position = get_option( 'edac_frontend_highlighter_position', 'right' );
+	?>
+		<fieldset>
+			<label>
+				<input type="radio" name="<?php echo 'edac_frontend_highlighter_position'; ?>" id="edac_frontend_highlighter_position" value="right" <?php checked( $position, 'right' ); ?>>
```
SteveJonesDev, line 328: Maybe the name attribute doesn't need to be PHP? If it does it should be escaped.

```suggestion
				<input type="radio" name="<?php echo esc_attr( 'edac_frontend_highlighter_position' ); ?>" id="edac_frontend_highlighter_position" value="right" <?php checked( $position, 'right' ); ?>>
```

SteveJonesDev (changes requested):

includes/options-page.php
```diff
@@ -285,6 +315,30 @@ function edac_simplified_summary_position_cb() {
...
+				<?php esc_html_e( 'Bottom Right Corner (default)', 'accessibility-checker' ); ?>
+			</label>
+			<br>
+			<label>
+				<input type="radio" name="edac_frontend_highlighter_position" value="left" <?php checked( $position, 'left' ); ?>>
```
SteveJonesDev, line 333: The two inputs' name attributes output should be consistent.

SteveJonesDev (changes requested):

includes/options-page.php
pattonwebz: Good catch, this doesn't need to be in PHP at all. Will inline it just. I copied this pattern over from other callbacks though. I will make a separate PR to fix those as well.

SteveJonesDev (approved):

PR #672, opened 2024-06-20
Summary panel a11y

### LINKED ISSUE #358: Can the star be a link with screen reader text?
Screenshot 2023-11-20 at 9 34 55 AM

This doesn't make sense to a screen reader user - it just says "star." Let's link it to the footnote so they have a way to jump to footnote. Add an aria-label to the link "About passed tests."

### LINKED ISSUE #359: Edit screen summary should be grouped in a list
Screenshot 2023-11-20 at 9 40 03 AM

This should be an unordered list. The numbers can be styled larger with spans, but each tile should be a list item. 

Also add an area-label on the li that has the count and label - I.e., aria-label="19 warnings" - because the different rows are confusing on a screen reader.

SteveJonesDev:
This PR updates the summary tiles to an unordered list with aria labels for improved a11y.

- Added helper class to output summary items as list items with aria labels.
- Update summary labels to be translatable.
- Update CSS logic
- Linked asterisk to description and added screen reader text

Fixes #359
Fixes #358

### DIFF
--- a/admin/class-ajax.php
+++ b/admin/class-ajax.php
@@ -100,54 +100,56 @@ class Ajax {
 			}
 		}
 
-		$html['content'] .= '<div class="edac-summary-total">';
+		$html['content'] .= '<ul>';
+		
+			$html['content'] .= '<li class="edac-summary-total" aria-label="' . $summary['passed_tests'] . '% Passed Tests">';
 
-		$html['content'] .= '<div class="edac-summary-total-progress-circle ' . ( ( $summary['passed_tests'] > 50 ) ? ' over50' : '' ) . '">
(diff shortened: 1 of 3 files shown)


pattonwebz (approved):

pattonwebz:
I tested this, and the summary works with it. There are no errors after checkout, as far as I can tell.

@SteveJonesDev I'm unsure if this is intentional, but the panels' background colour has changed with this PR.

Without PR:
Screenshot from 2024-06-24 18-38-47

After checkout and build this PR:
Screenshot from 2024-06-24 18-39-26

pattonwebz (approved):

PR #680, opened 2024-06-25
Release v1.14.0

pattonwebz:
Steve Jones (4):
- [x] Settings Styles
- [x] Updated view on page aria-label with issue ID
- [x] Added max width to issue image/svg
- [x] Summary panel a11y

William Patton (7):
- [x] Add a filter for the screens which remove admin notices
- [x] Add action in js validate that sends the request through to operate on
- [x] Allow frontend highlighter position to be moved from right to left of screen
- [x] Check all child nodes inside a paragraph rather than just the first
- [x] Add filter to allow frontend highlighter to be shown to other users
- [x] Add some handling in ignore method to support ignoring large batches …
- [x] Add some checks for namespace updates


readme.txt
```diff
@@ -3,7 +3,7 @@ Contributors: equalizedigital, alh0319, stevejonesdev
 Tags: accessibility, accessible, wcag, ada, WP accessibility
 Requires at least: 6.2
 Tested up to: 6.5.4
```
SteveJonesDev, line 5:
```suggestion
Tested up to: 6.5.5
```

SteveJonesDev (changes requested):

SteveJonesDev (approved):

PR #681, opened 2024-06-25
Wrap the summary items inside only a ul and li items (removing a div)

pattonwebz:
This PR removes a wrapper div, making summary items be `<li>` elements only so they are read out correctly to a screen reader.

The styles were tweaked to use CSS grid for styling. There are some minor differences after the style changes.

### DIFF
--- a/admin/class-ajax.php
+++ b/admin/class-ajax.php
@@ -100,8 +100,8 @@ class Ajax {
 			}
 		}
 
-		$html['content'] .= '<ul>';
-		
+		$html['content'] .= '<ul class="edac-summary-grid">';
+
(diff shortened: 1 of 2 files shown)


SteveJonesDev (changes requested):
The stats should stack sooner. They used to stack when the viewport went under 1140 pixels.

Screenshot 2024-06-25 at 2 56 46 PM

To prevent this layout from happening.
Screenshot 2024-06-25 at 2 59 13 PM

pattonwebz:
@SteveJonesDev I swapped it to use the lg breakpoint as the one that swaps it to the 2 col layout instead of the md breakpoint.

SteveJonesDev (approved):

Issue #688, opened 2024-06-26
Don't flag underlines that are inside a link


pattonwebz:
A user reported on dot org forums that the underlined link error was flagged on their site for a header - but the header was inside an anchor.

An item inside an anchor with an underline is valid and should not be flagged.

PR #689, opened 2024-06-26
William/688/prevent underlined text from detecting elements that are inside links checks 3 levels deep

pattonwebz:
This PR adds some checks in the 'matches' value of the underlined_text rule definition. The check are on the parent node (as high as 3 levels) for anchors and if it finds one won't check to flag an underline since flagging them if inside anchors is a false positive.

Closes: #688

### DIFF
--- a/src/pageScanner/rules/underlined-text.js
+++ b/src/pageScanner/rules/underlined-text.js
@@ -10,10 +10,28 @@ export default {
 	impact: 'moderate',
 	selector: '*:not(a):not(html):not(head):not(body)',
 	matches: ( element ) => {
-		return (
-			! element.hasAttribute( 'href' ) ||
-			element.tagName.toLowerCase() === 'u'
-		);
+		// U tags we will always check.
+		if ( element.tagName.toLowerCase() === 'u' ) {
+			return true;
+		}
+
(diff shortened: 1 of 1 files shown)


src/pageScanner/rules/underlined-text.js
```diff
@@ -8,11 +8,30 @@
...
+		// U tags we will always check.
+		if ( element.tagName.toLowerCase() === 'u' ) {
+			return true;
+		}
+
+		// Traverse up the dom 3 levels and check if the element is inside an anchor.
+		let parent = element.parentNode;
+		let isInsideAnchor = false;
+		for ( let i = 0; i < 3; i++ ) {
+			// can't go further up the dom if the parent is the html element.
```
pattonwebz, line 22:
```suggestion
			// Can't go further up the dom if the parent is the html element.
```

SteveJonesDev:
@pattonwebz, should this pass or fail?

`<a href="https://example.com"><u>This is a link with underline</u></a>`

pattonwebz:
> @pattonwebz, should this pass or fail?
> 
> `<a href="https://example.com"><u>This is a link with underline</u></a>`

Probably fail, all `u` tags should fail

SteveJonesDev (approved):

PR #692, opened 2024-06-28
Purge the post data if the saved post is in or is moving to trash

pattonwebz:
This PR calls the method to delete issues when a post is trashed during the validate method at the point it would early bail since post is trashed.

I moved the trash bail later in the method to ensure that any other quicker bail checks can run before it runs database queries for deleting issues.

This is needed here because Gutenberg editor does not fire the `wp_trash_post` hook because it uses rest API which updates a post with trash status where as classic editor and the quick actions in post lists invoke the actual trash functions. Deleting via rest does have a different action we could bind to - `rest_delete_{$post_type}` but using it is less clean than the way `wp_trash_post` works so I decided to use the method in this PR to solve this instead.

### DIFF
--- a/includes/validate.php
+++ b/includes/validate.php
@@ -94,6 +94,10 @@ function edac_save_post( $post_ID, $post, $update ) {
 
 	// Post in, or going to, trash.
 	if ( 'trash' === $post->post_status ) {
+		// Gutenberg does not fire the `wp_trash_post` action when moving posts to the
+		// trash. Instead it uses `rest_delete_{$post_type}` which passes a different shape
+		// so instead of hooking in there for every post type supported the data gets
+		// purged here instead which produces the same result.
 		Purge_Post_Data::delete_post( $post_ID );
 		return;
 	}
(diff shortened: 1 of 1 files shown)


includes/validate.php
```diff
@@ -96,6 +92,12 @@ function edac_save_post( $post_ID, $post, $update ) {
 		}
 	}
 
+	// Post in, or going to, trash.
+	if ( 'trash' === $post->post_status ) {
+		Purge_Post_Data::delete_post( $post_ID );
```
SteveJonesDev, line 97: Maybe add a comment.

SteveJonesDev (changes requested):

includes/validate.php
pattonwebz, line 97:
```suggestion
		// Gutenberg does not fire the `wp_trash_post` action when moving posts to the
		// trash. Instead it uses `rest_delete_{$post_type}` which passes a different shape
		// so instead of hooking in there for every post type supported the data gets
		// purged here instead which produces the same result.
		Purge_Post_Data::delete_post( $post_ID );
```

SteveJonesDev (approved):

Issue #598, opened 2024-05-01
Missing Transcript False Positives on .avif images


amberhinds:
<img width="1302" alt="Screenshot 2024-05-01 at 12 51 47 PM" src="https://github.com/equalizedigital/accessibility-checker/assets/13153456/8718e2e7-e49a-4702-a360-7087abcf78be">

* Accessibility Checker Version 1.10.2

This is on Dream A World.

pattonwebz:
I haven't looked closely at this, but my initial thought is that we need to not run video rules against these and instead treat `avif` file formats as images - like we treat gifs and web files that can be animated.

amberhinds:
Yeah, I had to google that file format, but it appears to only be used by images not videos.

pattonwebz:
I just looked at this, and the same false positive happens with .avif files being considered as videos, and `video_present` is flagged for them as well.

pattonwebz:
@amberhinds I have a solution for this in PR #695. It will prevent `.avif` being considered for checks that the file format isn't applicable to.

amberhinds:
Thank you!

PR #695, opened 2024-06-28
Fix: Don't flag .avif as missing trascript or video present

pattonwebz:
This PR adjusts how file extensions are checked in some rules so that `avif` does not get checked by them since it is not a video.

The reason this was happening was due to the way file extensions were being checked. They were matching `avi` in `avif`. I have created a helper that checks the extension matches the full extension and not just part of it.

Fixes: #598

### DIFF
--- a/includes/classes/class-edac-dom.php
+++ b/includes/classes/class-edac-dom.php
@@ -136,9 +136,7 @@ class EDAC_Dom extends simple_html_dom {
 			$elements = array_filter(
 				$elements_with_src,
 				function ( $element ) use ( $extensions ) {
-					$count = 0;
-					str_ireplace( $extensions, '', $element->getAttribute( 'src' ), $count );
-					return $count > 0;
+					return is_item_using_matching_extension( $element->getAttribute( 'href' ), $extensions );
 				}
 			);
 		}
--- a/includes/classes/class-edac-dom.php
+++ b/includes/classes/class-edac-dom.php
@@ -161,9 +161,7 @@ class EDAC_Dom extends simple_html_dom {
 			$elements = array_filter(
 				$elements_with_href,
 				function ( $element ) use ( $extensions ) {
-					$href      = $element->getAttribute( 'href' );
-					$extension = pathinfo( $href, PATHINFO_EXTENSION );
-					return in_array( '.' . $extension, $extensions, true );
+					return is_item_using_matching_extension( $element->getAttribute( 'href' ), $extensions );
 				}
 			);
 		}
--- a/includes/classes/class-edac-dom.php
+++ b/includes/classes/class-edac-dom.php
@@ -136,7 +136,7 @@ class EDAC_Dom extends simple_html_dom {
 			$elements = array_filter(
 				$elements_with_src,
 				function ( $element ) use ( $extensions ) {
-					return is_item_using_matching_extension( $element->getAttribute( 'href' ), $extensions );
+					return edac_is_item_using_matching_extension( $element->getAttribute( 'href' ), $extensions );
 				}
 			);
 		}
@@ -159,7 +159,7 @@ class EDAC_Dom extends simple_html_dom {
 			$elements = array_filter(
 				$elements_with_href,
 				function ( $element ) use ( $extensions ) {
-					return is_item_using_matching_extension( $element->getAttribute( 'href' ), $extensions );
+					return edac_is_item_using_matching_extension( $element->getAttribute( 'href' ), $extensions );
 				}
 			);
 		}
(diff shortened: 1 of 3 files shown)


includes/helper-functions.php
```diff
@@ -990,3 +990,18 @@ function edac_generate_summary_stat( string $item_class, int $count, string $lab
...
+ * Check if an element has an extension that matches the provided list.
+ *
+ * @since 1.15.0
+ *
+ * @param string $item A file path or URL to check.
+ * @param array  $extensions An array of extensions to check for.
+ *
+ * @return bool
+ */
+function is_item_using_matching_extension( string $item, array $extensions ): bool {
```
SteveJonesDev, line 1004:
```suggestion
function edac_is_item_using_matching_extension( string $item, array $extensions ): bool {
```

includes/helper-functions.php
```diff
@@ -990,3 +990,18 @@ function edac_generate_summary_stat( string $item_class, int $count, string $lab
...
+	$extension = pathinfo( $item, PATHINFO_EXTENSION );
+	return in_array( '.' . $extension, $extensions, true );
```
SteveJonesDev, line 1006: Are extensions always passed with the dot included?

Maybe we should always strip the dot from the extension to ensure passing with or without the dot works?

includes/helper-functions.php
```diff
@@ -990,3 +990,18 @@ function edac_generate_summary_stat( string $item_class, int $count, string $lab
...
+
+/**
+ * Check if an element has an extension that matches the provided list.
+ *
+ * @since 1.15.0
+ *
+ * @param string $item A file path or URL to check.
+ * @param array  $extensions An array of extensions to check for.
+ *
+ * @return bool
```
SteveJonesDev, line 1002:
```suggestion
 * @return bool True if the item's extension matches one of the provided extensions, false otherwise.
```

includes/helper-functions.php
SteveJonesDev, line 1004: Should we added some input validation?

```php
if ( ! is_string( $item ) || ! is_array( $extensions ) ) {
   return false;
}
```

SteveJonesDev (changes requested):

includes/helper-functions.php
pattonwebz: With pathinfo() and PATHINFO_EXTENSION, the extension is always omitted. I also wondered if we could strip the dot from the $extensions arrays, but I figured it could have some unintended consequences that would require more thorough testing.

includes/helper-functions.php
pattonwebz: The typed params mean that the values here can only ever be a string and an array, so it's not needed inside the function. This check could never fail and return early.

includes/rules/video_present.php
```diff
@@ -65,10 +65,8 @@ function edac_rule_video_present( $content, $post ) { // phpcs:ignore -- $post i
 				continue;
 			}
 
-			foreach ( $file_extensions as $file_extension ) {
-				if ( strpos( strtolower( $src_text ), $file_extension ) ) {
-					$errors[] = $element->outertext;
-				}
+			if ( is_item_using_matching_extension( $src_text, $file_extensions ) ) {
```
pattonwebz, line 68:
```suggestion
			if ( edac_is_item_using_matching_extension( $src_text, $file_extensions ) ) {
```

includes/classes/class-edac-dom.php
```diff
@@ -136,9 +136,7 @@ public function find_media_embeds( $include_audio = true ) {
 			$elements = array_filter(
 				$elements_with_src,
 				function ( $element ) use ( $extensions ) {
-					$count = 0;
-					str_ireplace( $extensions, '', $element->getAttribute( 'src' ), $count );
-					return $count > 0;
+					return is_item_using_matching_extension( $element->getAttribute( 'href' ), $extensions );
```
pattonwebz, line 139:
```suggestion
					return edac_is_item_using_matching_extension( $element->getAttribute( 'href' ), $extensions );
```

includes/classes/class-edac-dom.php
```diff
@@ -161,9 +159,7 @@ public function find_linked_media( $include_audio = true ) {
 			$elements = array_filter(
 				$elements_with_href,
 				function ( $element ) use ( $extensions ) {
-					$href      = $element->getAttribute( 'href' );
-					$extension = pathinfo( $href, PATHINFO_EXTENSION );
-					return in_array( '.' . $extension, $extensions, true );
+					return is_item_using_matching_extension( $element->getAttribute( 'href' ), $extensions );
```
pattonwebz, line 162:
```suggestion
					return edac_is_item_using_matching_extension( $element->getAttribute( 'href' ), $extensions );
```

SteveJonesDev (approved):

Issue #610, opened 2024-05-05
Missing Form Label should not flag on image input with alt text


amberhinds:
https://github.com/equalizedigital/accessibility-checker/blob/183f257855902a96deb54e1b1d51688bfafed9fa/includes/rules/missing_form_label.php#L15

[See post on rule testing site.](https://a11ycheckrules.wpenginepowered.com/missing-form-label/)

If it's an input with a type of image AND alt text, then I would not expect this to flag.

pattonwebz:
@amberhinds I worked though this in #696 and it should be in the next release.

amberhinds:
Thanks!

PR #696, opened 2024-06-28
Enahcement: Image inputs with alt text shouldn't flag for missing_form_label

pattonwebz:
This PR adds a check inside of the missing_form_label that will prevent an error being flagged for an input with type of 'image' if it has non-empty alt text. A screen reader will use the alt text even when there is no label.

Fixes: #610


SteveJonesDev (approved):
This works in my testing.

Before:
Screenshot 2024-07-01 at 3 38 35 PM

After:
Screenshot 2024-07-01 at 3 39 21 PM

Issue #697, opened 2024-07-01
Empty link rule check misses empty links


SteveJonesDev:
The `empty_link` check misses empty links. Issue reproduced using this test code `<a href="https://example.com"></a>`.

@pattonwebz is this bug caught when running the end-to-end test in Cypress?

Let's do a patch release on this given its impact.

Ticket: https://wordpress.org/support/topic/cannot-get-empty-link-error-to-trigger/

oldrup:
Looking forward to testing the patch. @SteveJonesDev, what's the preferred channel to get in touch if experiencing potential issues with the free plugin, here on GitHub or the WordPress forums?

SteveJonesDev:
@oldrup either channel is fine.

pattonwebz:
I have a PR up #699 that should fix this and adds some tests to verify that it is working as expected.

@oldrup thanks for reporting this, let me know if you want me to send you a test build to verify this is resolved for you before I release this to the plugin repo.

PR #699, opened 2024-07-01
Fixed: Allow empty_link rule to detect actually empty links

pattonwebz:
This PR makes is so that empty links are flagged appropriately by:

* Correcting some variables that were being checked
* Flagging an error if it's empty plain text and there is no `img`, `input` or `i` tags for further checks

Tests were also added to confirm the behaviour of the rule matches the intention against some test data.

Fixes: #697


SteveJonesDev (approved):
LGTM in my testing. Thanks for adding the Unit Tests.

Issue #694, opened 2024-06-28
Customer reported contrast fail on CTA button hover state in the ACP upsell on our settings page 


christophermhinds:
I did not follow the standard bug report format because I believe this screenshot should be sufficient.

<img width="292" alt="Screenshot 2024-06-28 at 8 43 37 AM" src="https://github.com/equalizedigital/accessibility-checker/assets/25856499/0c53d37e-22fc-4ce2-9203-c57cb97d79a5">

Happy to re-submit if necessary.

pattonwebz:
@christophermhinds can you ask this customer what browser and version of it they are using? I see something different when I spin up a fresh site for this button. It looks like what they are seeing is backwards compared to what I see.

https://github.com/equalizedigital/accessibility-checker/assets/3902039/e4bfd6f9-4a09-46a3-b38a-0ef7cad37e7d

tdrayson:
Hey @pattonwebz, I'm using FireFox (developer edition), version 128.

I found removing `z-index: 1;` from the `.edac-pro-callout-button` https://github.com/equalizedigital/accessibility-checker/blob/8b602e744374add199c79bb7c00c3c2ba6e70d7b/src/admin/sass/accessibility-checker-admin.scss#L1025

and adding `z-index: 0` to `edac-pro-callout` https://github.com/equalizedigital/accessibility-checker/blob/8b602e744374add199c79bb7c00c3c2ba6e70d7b/src/admin/sass/accessibility-checker-admin.scss#L973

This seems to fix the issue.

pattonwebz:
@tdrayson thanks for sharing the browser version you are using. When I tested with FF I found a similar fix worked as well but to figure out why I dug a little deeper and discovered that the difference between what is seen in FF and what I seen in my Chrome when I first looked comes down to differences in how they handle stacking contexts.

Firefox appears to be the only browser that I tested that implements stacking contexts exactly to specification and other browsers have a more loose implementation.

I'll have a solution pushed up tomorrow, will be released in the next plugin release.

Appreciate you reporting this and bringing it to our attention, this was interesting for me to dig into.

tdrayson:
Perfect, thanks for the investigation. Always fun when different browsers implement specifications differently.... 🤣

PR #704, opened 2024-07-08
Wraps the pro callout button and tweak z-index to handle stacking contexts

pattonwebz:
This PR updates the index around pro callout buttons so that the stacking context is appropriately handled across various browsers. It fixes a contrast issue it caused on Firefox.

* Adds a wrapper around the button with a z-index to create a new stacking context.
* Removes z-index on the actual button to avoid creating an additional context.
* Makes the button text inline-block to avoid a broken button split.

With these changes it is now fixed in Firefox (and looks no different in other browsers I tested)
https://github.com/equalizedigital/accessibility-checker/assets/3902039/8cec5d3c-156f-4270-9bdb-59dacc35f098

Fixes: #694


SteveJonesDev (changes requested):
The spacing has shifted. See the screenshot, original on the left and modified on the right.
Screenshot 2024-07-09 at 10 03 06 AM

Did you mean to say "it is NOW fixed in Firefox"?

> With changes it is not fixed in Firefox (and looks no different in other browsers I tested)

pattonwebz:
@SteveJonesDev, I swapped to an inline display for the button again, and it looks the same as it does without these CSS tweaks. I also added a Firefox-specific padding tweak to prevent a size/padding difference in that browser. The base font display is slightly larger in FF, and that causes the button to wrap to 2 lines.

https://github.com/equalizedigital/accessibility-checker/assets/3902039/445bb8f6-fac9-40ac-a39b-37573156ea11

SteveJonesDev (approved):

Issue #386, opened 2023-12-07
Too many single-use variables


jdevalk:
[This code](https://github.com/equalizedigital/accessibility-checker/blob/22a765d3cf085aea193039b5bec322d827dd2d08/includes/insert.php#L117): 
```php
$error = new WP_Error( '-2', 'The post ID was not set' );
wp_send_json_error( $error );
```
could be:
```php
wp_send_json_error(
    new WP_Error( '-2', esc_html__( 'The post ID was not set', 'accessibility-checker' ) )
);
```
No setting variables when you don't need to, properly translating text strings, etc.

PR #710, opened 2024-07-10
Inline some single use variables at function returns

pattonwebz:
This PR inlines variables at the return statement if they existed only to be returned. None of these seemed to aid in readability in my view so I swapped every instance that was detected.

There were 2 other instances inside the `simple_html_dom.php` file that I did not modify here since that package is a cloned 3rd party file.

### DIFF
--- a/admin/class-helpers.php
+++ b/admin/class-helpers.php
@@ -120,9 +120,7 @@ class Helpers {
 			}
 		}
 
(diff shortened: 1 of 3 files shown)


SteveJonesDev:
Partially fixes #386

SteveJonesDev (approved):

PR #715, opened 2024-07-15
Release v1.15.0

pattonwebz:
William Patton (11):
- [x] Enhancement: Image inputs with alt text shouldn't flag for missing_form_label
- [x] Minor cleanup of admin styles
- [x] Fix: Don't flag .avif as missing transcript or video present
- [x] Purge the post data if the saved post is in or is moving to the trash
- [x] Wraps the pro callout button and tweak z-index to handle stacking contexts
- [x] Add PHPStan to the project
- [x] Inline some single-use variables at function returns
- [x] Add a retry logic on some flaky tests that rely on httpbin
- [x] Add wp cli commands to get stats and delete stats
- [x] Add a basic PR template
- [x] PHP 8.4 deprecation notice fix for implicitly nullable Meta_Boxes


readme.txt
```diff
@@ -2,8 +2,8 @@
 Contributors: equalizedigital, alh0319, stevejonesdev
 Tags: accessibility, accessible, wcag, ada, WP accessibility
 Requires at least: 6.2
-Tested up to: 6.5.5
-Stable tag: 1.14.3
+Tested up to: 6.6
```
SteveJonesDev, line 5: Should this be semantic versioning?

```suggestion
Tested up to: 6.6.0
```

SteveJonesDev (changes requested):

readme.txt
pattonwebz: @SteveJonesDev I updated this with the 3 digit but I remember looking at the parser for readmes a while back and I think it ignores the last digit anyways. No harm having it but I don't think it checks it when doing the routine that shows the 'not tested' check.

SteveJonesDev: I looked at other plugins and I can confirm it updates the patch versions automatically.

SteveJonesDev (approved):

PR #725, opened 2024-08-02
Add framework to handle various different fixes that the plugin can apply

pattonwebz:
This PR introduces a framework to manage various fixes that the plugin can apply to users sites. There are no actual fixes added yet in this PR as it just puts the framework in place for adding the fixes.

- A settings page is added, initially supporting just checkbox type options
- Filters exist to add sections to the options page and register settings to any section on it
- Actions (will be) added where the fixes will have to be run or triggered.

## Checklist

- [x] PR is linked to a card

### DIFF
--- a/accessibility-checker.php
+++ b/accessibility-checker.php
@@ -20,7 +20,6 @@
  */
 
 use EDAC\Inc\Plugin;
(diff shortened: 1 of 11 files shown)


SteveJonesDev (changes requested):
Just one question.

SteveJonesDev (approved):

PR #734, opened 2024-09-03
Fix: add lang and dir attributes to html element

pattonwebz:
This PR adds a _fix_ that users can toggle on to include the `lang` and `dir` attributes on the root `html` element.

The fix uses a filter on `language_attribute` to add them but includes an additional JS fallback for themes which do not have the `language_attributes()` function in the correct place.

Note: This is based from the skip-link fix which is why the diff currently looks so large. It does also include some changes to how the fixes manager registers and invokes fixes - namely the inclusion of a `run` method requirement inside of fixes to separate settings and registration code from actual runner code.

## Checklist


SteveJonesDev (approved):
LGTM

PR #737, opened 2024-09-04
Add a fix class for adding post title to read more links

pattonwebz:
This adds it to the normal read more, the excerpt and the excerpt more.

Note: This does not work with the query loop block as they explicitly override anything from the filter if someone has filled in the read more text element in the block settings. See: https://github.com/WordPress/gutenberg/blob/abf72c95b1da1b8f6709858bb750e1d56782c27d/packages/block-library/src/post-excerpt/index.php#L39C5-L47C5

## Checklist

### DIFF
--- a/includes/classes/Fixes/Fix/ReadMoreAddTitleFix.php
+++ b/includes/classes/Fixes/Fix/ReadMoreAddTitleFix.php
@@ -108,6 +108,11 @@ class ReadMoreAddTitleFix implements FixInterface {
 
 		global $id;
 
+		// If the $text already contains the title, we won't add it again.
+		if ( str_contains( strtolower( $text ), strtolower( get_the_title( $id ) ) ) ) {
+			return $link;
+		}
+
 		return str_replace(
 			$text,
 			$text . ' ' . $this->generate_read_more_string( $id ),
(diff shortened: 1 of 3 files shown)


SteveJonesDev (changes requested):
Can we check if the title already exists so we don't get duplicate screen reader text?
```html
<a href="http://accessibilitychecker.local/template-sticky/#more-1241" class="more-link">
  <span class="faux-button">
    Continue reading
    <span class="edac-content-more-title edac-screen-reader-text">
      Template: Sticky
    </span>
  </span>
  <span class="screen-reader-text">“Template: Sticky”</span>
</a>

```

pattonwebz:
@SteveJonesDev not all of these actions receive the text content of the thing they are operating on so it isn't possible to do a direct check for post title in them all. I have added a check where I could but it won't be able to catch every time this could happen.

It's likely that some themes will add this in theme templates directly instead of ways that use the filters so it won't be possible to be 100% certain with catching. Because of that I have added a warning notice to settings group that says: `This setting is not recommended for themes that are already accessibility-ready.`

SteveJonesDev (approved):
Good additional checks!

PR #740, opened 2024-09-04
added: force underline fix

SteveJonesDev:
This PR adds the Force Underline on links Fix.

1. Excludes links inside of `nav` elements
2. Add filter `edac_filter_frontend_fixes_data` to allow for changing the target elements. The default is `a`.

Card: https://3.basecamp.com/3579237/buckets/29333825/card_tables/cards/7652429605

### DIFF
--- a/src/frontendFixes/Fixes/underlineFix.js
+++ b/src/frontendFixes/Fixes/underlineFix.js
@@ -1,0 +1,50 @@
+const ForceUnderlineFixData = window.edac_frontend_fixes?.underline || {
+	enabled: false,
+};
+
+const ForceUnderlineFix = () => {
+	if ( ! ForceUnderlineFixData.enabled ) {
+		return;
+	}
+
+	// Apply underline to any link that is not within a `nav` element.
+	// Using JavaScript ensures this style takes precedence over conflicting CSS rules.
+	const targets = document.querySelectorAll( ForceUnderlineFixData.target );
+	const originalOutline = targets[ 0 ].style.outlineWidth;
+	const originalOffset = targets[ 0 ].style.outlineOffset;
+	const textColor = targets[ 0 ].style.color;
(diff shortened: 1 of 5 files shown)


src/frontendFixes/Fixes/underlineFix.js
```diff
@@ -0,0 +1,50 @@
...
+
+		target.addEventListener( 'mouseleave', function() {
+			target.style.textDecoration = 'underline';
+		} );
+
+		target.addEventListener( 'focusin', function() {
+			let newOutline = '2px';
+			if ( originalOutline === '2px' ) {
+				newOutline = '4px';
+			}
```
pattonwebz, line 35: This is a nice solution to an annoying edge case!

src/frontendFixes/Fixes/underlineFix.js
```diff
@@ -0,0 +1,50 @@
...
+		return;
+	}
+
+	// Apply underline to any link that is not within a `nav` element.
+	// Using JavaScript ensures this style takes precedence over conflicting CSS rules.
+	const targets = document.querySelectorAll( ForceUnderlineFixData.target );
+	const originalOutline = targets[ 0 ].style.outlineWidth;
+	const originalOffset = targets[ 0 ].style.outlineOffset;
+	const textColor = targets[ 0 ].style.color;
+	const originalColor = targets[ 0 ].style.outlineColor;
```
pattonwebz, line 16: Would storing these originals on each element be better in the `focusin` event? Maybe put them in `data-*` attributes when swapped so that each targets own original styles could be restored?

SteveJonesDev: Yeah, this would be better. I've pushed an update.

pattonwebz (approved):

PR #742, opened 2024-09-05
Added: Fix for unscalable meta viewport tags

pattonwebz:
Adds a fix that will ensure the meta viewport tag does not prevent user scaling.

## Checklist

### DIFF
--- a/src/frontendFixes/Fixes/metaViewportScalableFix.js
+++ b/src/frontendFixes/Fixes/metaViewportScalableFix.js
@@ -11,12 +11,11 @@ const MetaViewportScalableFix = () => {
 	const metaViewport = document.querySelector( 'meta[name="viewport"]' );
 	if ( metaViewport ) {
 		// check if it has scalable set to no.
-		if ( metaViewport.content.includes( 'user-scalable=no' ) ) {
-			// Remove the meta viewport tag to allow it to be recreated.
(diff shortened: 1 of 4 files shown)


src/frontendFixes/Fixes/metaViewportScalableFix.js
```diff
@@ -0,0 +1,29 @@
...
+	if ( ! MetaViewportScalable.enabled ) {
+		return;
+	}
+
+	// Get the meta viewport tag.
+	const metaViewport = document.querySelector( 'meta[name="viewport"]' );
+	if ( metaViewport ) {
+		// check if it has scalable set to no.
+		if ( metaViewport.content.includes( 'user-scalable=no' ) ) {
+			// Remove the meta viewport tag to allow it to be recreated.
```
SteveJonesDev, line 15: Isn't this missing?

`metaViewport.remove();`

SteveJonesDev (changes requested):

src/frontendFixes/Fixes/metaViewportScalableFix.js
pattonwebz: Yes, this was missing. I refactored this to add it and avoid an if..else

SteveJonesDev (approved):

PR #746, opened 2024-09-10
Focus Outline Fix

SteveJonesDev:
This PR adds a fix to ensure a visible focus outline is applied to all focusable elements.

- Added: `focus_outline` settings
- Added: inline `<style>` tag to output CSS. We'll circle back and add a pattern for adding custom CSS.
- Added: a new field for color inputs
- Added: a new parameter called `default` to the `edac_filter_fixes_settings_fields` filter. Default logic will need to be added to the other field types if needed.
- Fixed: settings table `tbody` display so it spans the full widget of it's container


pattonwebz (approved):
The code here looks good to me. I tested the PR locally and can see focus outline added to elements with the color that I selected (and a white outline as well).

PR #750, opened 2024-09-11
Delete fix option keys on uninstall

pattonwebz:
This PR stores options keys so they can be later deleted if the plugin gets uninstalled

## Checklist

### DIFF
--- a/admin/AdminPage/FixesPage.php
+++ b/admin/AdminPage/FixesPage.php
@@ -154,7 +154,11 @@ class FixesPage implements PageInterface {
 		 */
 		$fields = apply_filters( 'edac_filter_fixes_settings_fields', [] );
 
+		$field_keys = [];
+
 		foreach ( $fields as $field_id => $field ) {
+
+			$field_keys[] = $field_id;
 
 			$field_type = $field['type'] ?? 'checkbox';
 			$sanitizer  = $field['sanitize_callback'] ?? [ $this, 'sanitize_checkbox' ];
@@ -176,6 +180,8 @@ class FixesPage implements PageInterface {
 
 			register_setting( self::SETTINGS_SLUG, $field_id, $sanitizer );
 		}
+
+		update_option( 'edac_fixes_fields', $field_keys );
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


SteveJonesDev:
@pattonwebz is this ready for code review?

pattonwebz:
@SteveJonesDev let me know what you think of the method used here. It feels janky but it does work. If you also think it's janky I will rethink it.

SteveJonesDev:
At first look, it seems a bit odd. Is that update_option going to get called on every page load? We can't just return an array?

pattonwebz:
Yeah, it gets called on all admin page loads. That's my main issue with it as well.

I didn't want to have to define an array of option keys manually because we are likely to forget to update them when we make changes. There's somewhere around 20+ options currently with the fixes so far.

SteveJonesDev:
We could create a method in the class:

```php
<?php
public function get_option_field_slugs() {
    // Initialize an empty array for the slugs
    $slugs = [];

    // Apply the same filter to get the fields array, ensuring external modifications are included
    $fields = apply_filters('edac_filter_fixes_settings_fields', []);

    // Iterate through the fields to extract the slugs
    foreach ($fields as $field_id => $field) {
        // Add the field ID (slug) to the slugs array
        $slugs[] = $field_id;
    }

    // Return the array of slugs
    return $slugs;
}
```

and then output it in the uninstall file:

```php
<?php
// Step 1: Include the FixesPage class if necessary
require_once 'path/to/FixesPage.php'; // Adjust the path as necessary

// Step 2: Instantiate FixesPage
// Assuming 'manage_options' is a suitable capability for your use case
$fixesPage = new EqualizeDigital\AccessibilityChecker\Admin\AdminPage\FixesPage('manage_options');

// Step 3: Call the method
$optionFieldSlugs = $fixesPage->get_option_field_slugs();

// Use $optionFieldSlugs as needed
var_dump($optionFieldSlugs); // Just an example of usage
```

But I don't love this either.

I actually think we just manually add the options like we have for everything else. Keep it simple.

pattonwebz:
Yeah I think I am leaning towards just the array of options method because we can't run code from the plugin inside the uninstall hook since it's already deactivated by that point

SteveJonesDev (approved):
LGTM

PR #751, opened 2024-09-11
Set pattern for including css in fixes

pattonwebz:
This PR adds a root stylesheet for fixes that gets included which individual fixes can use to target for wp_add_inline_style when they need to include small snippets of CSS for the fix.

## Checklist


pattonwebz:
At this point so much time has passed on this PR that it's likely not worth doing anymore or not worth doing right now. The pattern here is fairly straightforward, so if we need to do this again in future, the same pattern will likely be found again and reused.

PR #764, opened 2024-10-02
Move fix setting to the settings page in a tab

SteveJonesDev:
This PR removes the Accessibility Fixes link from the plugin menu and moves it to a tab on the settings page.

- removed the menu link
- added tab via existing filters
- updated `aria-label` on tabs nav element

Fixes: https://3.basecamp.com/3579237/buckets/29333825/card_tables/cards/7810260579


pattonwebz (approved):
I tested this and the page is moved to a tab, there are no notices from it in my debug log with the change

PR #765, opened 2024-10-03
updated: missing_lang_attr rule to use axe core rules

SteveJonesDev:
This PR updates the `missing_lang_attr` rule to use Axe-core rules `html-lang-valid` and `html-has-lang`.

- Defined axe-core rules
- Update rules.php with `ruleset`, `combines`, and set `viewable` to false.
- Deleted php rule `missing_lang_attr.php`

Fixes: https://3.basecamp.com/3579237/buckets/29333825/card_tables/cards/7879186755


pattonwebz (approved):
I tested this and it works well for me.

I added a commit that strips the affected code snippet that gets saved down to be use the opening and closing tag to avoid storing the entire page in the database as the fragment. The issue can't be viewed on the frontend so no concern with finding this markup again when visiting there.

https://github.com/equalizedigital/accessibility-checker/pull/765/commits/814b901a04b13f028834ba9ad38f1f6cb50778bd

Screenshot from 2024-10-07 14-27-47

PR #766, opened 2024-10-03
updated: tab_order_modified rule to use axe-core

SteveJonesDev:
The PR updates the `tab_order_modified` rule to use the axe-core `tabindex` rule.

- Defined axe-core rule
- Update rules.php with ruleset and combines values
- Deleted php rule tabindex.php

Fixes: https://3.basecamp.com/3579237/buckets/29333825/card_tables/cards/7879201141


pattonwebz (approved):
I tested this PR with an element with a tabindex. It detects it correctly, when the fix to remove tab indexes is enabled the issue is no longer detected.

PR #768, opened 2024-10-04
updated: missing_form_label and empty_form_label to use axe-core label rule

SteveJonesDev:
This PR updates the `missing_form_label` rule to use Axe-core `rule` label. This Axe-core rule also checks for empty form labels so the old `empty_form_label` has been removed.

- Defined axe-core rule
- Update rules.php with ruleset and combines.
- Deleted php rule `missing_title.php` and `empty_form_label`

Fixes: https://3.basecamp.com/3579237/buckets/29333825/card_tables/cards/7879179791


SteveJonesDev:
@pattonwebz, PR looks good.

PR #771, opened 2024-10-07
Convert the `link_blank` rule to a js rule

pattonwebz:
Converts the php `link_blank` rule to a js rule to run in axe-core. This should be functionally the same (to be verified on test site).

Basecamp card: https://3.basecamp.com/3579237/buckets/29333825/card_tables/cards/7879174480

## Checklist


SteveJonesDev (changes requested):
See changes in Basecamp

SteveJonesDev (approved):

PR #772, opened 2024-10-08
updated: broken skip anchor link to JavaScript

SteveJonesDev:
This PR updates the `broken_skip_anchor_link` rule to use a custom axe-core rule.

- Defined custom axe-core rule and check
- Update rules.php with ruleset value
- Deleted php rule `broken_skip_anchor_link.php` and empty_form_label

Fixes: https://3.basecamp.com/3579237/buckets/29333825/card_tables/cards/7879202645


pattonwebz (approved):
- Tested with a skip link that points to a non-existent anchor and error is raised.
- Tested with a valid skip link and no error is raised
- Tested with no skip link and no error is raised.

PR #773, opened 2024-10-08
Don't flag an issue for empty paragraph if it has aria-hidden

pattonwebz:
Someone opened a support ticket with a reason they wanted empty paragraphs with aria-hidden to not flag and it seems like it's worth it since if they have aria-hidden then screen readers will ignore them anyway and avoid creating noise to a screen reader user.

You can test this with:

```html
<p></p>
<p aria-hidden="true"></p>
<p aria-hidden="false"></p>
<p aria-hidden="TRUE"></p>
```

## Checklist


SteveJonesDev (approved):
LGTM

PR #777, opened 2024-10-10
Exclude the query monitor container from the JS scan.

pattonwebz:
With some additional rules moved over the the JS scan some items inside the query monitor container were being flagged (new link warnings). 

This PR excludes the container from the scan by using the `context.exclude[]` object passed to the runner.

Excluded by id #query-monitor-main

## Checklist


SteveJonesDev (approved):
LGTM

PR #787, opened 2024-10-22
William/7652430604/attach correct fix to missing title rule

pattonwebz:
Attaches the correct fix to the missing_title rule so it shows the correct fields when prompted to fix in highlighter and issue detail panel.

It also strips down the stored markup to just the `<html>...</html>`element to prevent storing the entire page plus translates a string that was missed before.

Basecamp Cards:
* https://3.basecamp.com/3579237/buckets/29333825/card_tables/cards/7879198801
* https://3.basecamp.com/3579237/buckets/29333825/card_tables/cards/7652430604

## Checklist


SteveJonesDev (approved):
LGTM

PR #817, opened 2024-11-19
Show a notice when fixes settings are saved on options page

pattonwebz:
When fixes settings are saved, a transient is set to be used as a flag for notifications. It is used in the fixes_page.php partial to output a notice about rescanning and flushing the cache.

It's done this way rather than via admin_notifications because we are unhooking all other notifications when on the settings pages.

Screenshot from 2024-11-18 21-49-25

## Checklist

### DIFF
--- a/partials/admin-page/fixes-page.php
+++ b/partials/admin-page/fixes-page.php
@@ -7,7 +7,14 @@
 
 ?>
 <div id="edac-fixes-page" class="wrap edac-settings <?php echo EDAC_KEY_VALID ? '' : 'pro-callout-wrapper'; ?>">
-
+	<?php if ( get_transient( 'edac_fixes_settings_saved' ) ) : ?>
+		<div class="notice notice-warning">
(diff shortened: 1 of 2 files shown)


SteveJonesDev (changes requested):

partials/admin-page/fixes-page.php
```diff
@@ -7,7 +7,14 @@
 
 ?>
 <div id="edac-fixes-page" class="wrap edac-settings <?php echo EDAC_KEY_VALID ? '' : 'pro-callout-wrapper'; ?>">
-
+	<?php if ( get_transient( 'edac_fixes_settings_saved' ) ) : ?>
+		<div class="notice notice-warning is-dismissible">
+			<p><?php esc_html_e( 'Settings Saved. To confirm and reflect these updates in your site's statistics, please rescan your site. If you are using a caching plugin, make sure to clear the cache to ensure the updated fixes are applied correctly.', 'accessibility-checker' ); ?></p>
```
pattonwebz, line 12:
```suggestion
			<p><?php esc_html_e( "Settings Saved. To confirm and reflect these updates in your site's statistics, please rescan your site. If you are using a caching plugin, make sure to clear the cache to ensure the updated fixes are applied correctly.", 'accessibility-checker' ); ?></p>
```

SteveJonesDev (approved):

PR #819, opened 2024-11-21
Replace node sass with sass and

pattonwebz:
This PR replaces node-sass with the sass package and deals with some other deprecations issues and warnings:

* Replaces inline divisions with calc calls
* Replaces `@import` with `@use` statements
* Updates webpack and sass-loader to avoid legacy JS warning
* Tackles a mixed-decl issue with new generation order for styles that appear after other nested rules

## Checklist


SteveJonesDev (approved):
`npm install` is working with these changes.

PR #820, opened 2024-11-22
Fix: Improve table header detection logic to correctly validate row headers

SteveJonesDev:
This pull request includes updates to the `includes/rules/missing_table_header.php` file to improve the accessibility checks for table headers. The most significant changes involve renaming a function for clarity, modifying the logic to check for proper table headers, and ensuring that the function returns an appropriate value when no tables are found.

Key changes:

-  Renamed the function `edac_th_match_td` to `edac_has_proper_headers` for better clarity.
-  Enhanced the logic within the `edac_has_proper_headers` function to check for valid table headers by verifying the presence of header text and valid scope attributes.
-  Modified the `edac_rule_missing_table_header` function to return an empty errors array if no tables are found, ensuring consistency in the return type.
-  Corrected a typo in the file header comment from "pluign" to "plugin".

Fixes: https://3.basecamp.com/3579237/buckets/29333825/card_tables/cards/8057688694

### DIFF
--- a/includes/rules/missing_table_header.php
+++ b/includes/rules/missing_table_header.php
@@ -1,6 +1,6 @@
 <?php // phpcs:ignore WordPress.Files.FileName.NotHyphenatedLowercase -- underscore is for valid function name.
 /**
- * Accessibility Checker pluign file.
+ * Accessibility Checker plugin file.
  *
  * @package Accessibility_Checker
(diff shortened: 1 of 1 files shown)


pattonwebz (approved):
Tested this and it appears to work how it was intended.

PR #829, opened 2024-12-09
Enhancement: Retry frontend highlighter init later

pattonwebz:
This PR adds a simple retry later to init the Frontend Highlighter. This should solve the issue where Cloudflare Rocket Loader defer prevents the highlighter from initialising.

It will try init first on `DOMContentLoaded` and then later at `load`. It adds a variable as a flag to track if the init code has already run to prevent it duplicating on `load` if it already ran when initially intended.

## Checklist

### DIFF
--- a/src/frontendHighlighterApp/index.js
+++ b/src/frontendHighlighterApp/index.js
@@ -907,3 +907,7 @@ const initHighlighter = () => {
 window.addEventListener( 'DOMContentLoaded', () => {
 	initHighlighter();
 } );
+// Add a listener later than DOMCOntentLoaded to check if the init has run.
+window.addEventListener( 'load', () => {
+	initHighlighter();
+} );
--- a/src/frontendHighlighterApp/index.js
+++ b/src/frontendHighlighterApp/index.js
@@ -894,14 +894,14 @@ class AccessibilityCheckerHighlight {
 // Some systems (Cloudflare Rocket Loader) defers scripts for performance but that can
 // cause some DOMContentLoaded events to be missed. This is flag tracks if it run so we
 // can retry at a latter event listener.
-let initHasRun = false;
+let highlighterInitialized = false;
 const initHighlighter = () => {
-	if ( ! initHasRun ) {
+	if ( ! highlighterInitialized ) {
 		new AccessibilityCheckerHighlight();
 		if ( window.edacFrontendHighlighterApp?.userCanFix ) {
 			fixSettingsModalInit();
 		}
-		initHasRun = true;
+		highlighterInitialized = true;
 	}
 };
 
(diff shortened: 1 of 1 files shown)


src/frontendHighlighterApp/index.js
```diff
@@ -891,9 +891,23 @@ class AccessibilityCheckerHighlight {
...
 }
 
-window.addEventListener( 'DOMContentLoaded', () => {
-	new AccessibilityCheckerHighlight();
-	if ( window.edacFrontendHighlighterApp?.userCanFix ) {
-		fixSettingsModalInit();
+// Some systems (Cloudflare Rocket Loader) defers scripts for performance but that can
+// cause some DOMContentLoaded events to be missed. This is flag tracks if it run so we
+// can retry at a latter event listener.
+let initHasRun = false;
```
SteveJonesDev, line 897: I'm curious if this can be simplified by using the once parameter and combining the listeners.

```js
const initHighlighter = () => {
    new AccessibilityCheckerHighlight();
    if (window.edacFrontendHighlighterApp?.userCanFix) {
        fixSettingsModalInit();
    }
};

// Combine listeners for DOMContentLoaded and load, using `once` to ensure the handler runs only once
['DOMContentLoaded', 'load'].forEach(event => {
    window.addEventListener(event, initHighlighter, { once: true });
});
```

SteveJonesDev (changes requested):

src/frontendHighlighterApp/index.js
pattonwebz: I used an array loop for the listener events but the `once` prop  for `addEventListener` is once per listener and not once across all of them :(

SteveJonesDev (approved):
A minor improvement could be to change `initHasRun` to `highlighterInitialized` so it's more descriptive. Either way is fine, I'll let you decide.

src/frontendHighlighterApp/index.js
```diff
@@ -891,9 +891,20 @@ class AccessibilityCheckerHighlight {
...
+const initHighlighter = () => {
+	if ( ! initHasRun ) {
+		new AccessibilityCheckerHighlight();
+		if ( window.edacFrontendHighlighterApp?.userCanFix ) {
+			fixSettingsModalInit();
+		}
+		initHasRun = true;
 	}
+};
```
pattonwebz, line 906:
```suggestion
let highlighterInitialized = false;
const initHighlighter = () => {
	if ( ! highlighterInitialized ) {
		new AccessibilityCheckerHighlight();
		if ( window.edacFrontendHighlighterApp?.userCanFix ) {
			fixSettingsModalInit();
		}
		highlighterInitialized = true;
	}
};
```

PR #836, opened 2025-01-16
Make the endpoint for clearing issues less strict

pattonwebz:
This now allows posts that no longer exist to be flushed out of the table of issues when passed a specific 'skip_post_exists_check' parameter.

Why: A support request came in about issues existing for a post that was deleted. It is difficult to clear these issues out without database access and the clear-issues endpoint had a check to validate that the post exists in it and that the post type was valid. 

## Checklist

### DIFF
--- a/includes/classes/class-rest-api.php
+++ b/includes/classes/class-rest-api.php
@@ -231,14 +231,16 @@ class REST_Api {
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


SteveJonesDev (approved):
Looks good.

PR #842, opened 2025-01-24
Prevent password protected notice being triggered by woocommerse checkout page scan

pattonwebz:
This PR detects when a scan fail occurs on the woocommerse checkout page (due to a redirect when cart is empty) and returns early before the `edac_password_protected` option is set.

The page still will not scan after this change - but it will not show an inaccurate notice when it occurs.

We will need to cycle back in future to resolve actually being able to scan the checkout page in different situations.

## Checklist

## Summary by CodeRabbit

- **New Features**
	- Added helper functions to check WooCommerce plugin status and identify the checkout page.
	- Enhanced validation process for WooCommerce checkout pages.

- **Bug Fixes**
	- Resolved an issue with content validation on WooCommerce checkout pages.
	- Improved handling of empty content scenarios during page validation.

### DIFF
--- a/includes/helper-functions.php
+++ b/includes/helper-functions.php
@@ -1047,7 +1047,7 @@ function edac_generate_link_type( $query_args = [], $type = 'pro', $args = [] ):
  *
  * @return bool
  */
-function edac_is_woocommerse_enabled() {
+function edac_is_woocommerce_enabled() {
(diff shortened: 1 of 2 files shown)


SteveJonesDev (approved):
Looks good.

PR #845, opened 2025-01-31
Rework urls counts and values of post scanned and other welcome page stats

pattonwebz:
The main change in this PR is swapping the `posts_scanned` value used in several places and calculations to actually contain a number that is the total number of posts scanned. The number previously contained a value that was 'posts_with_issues`.

* posts_scanned value now is generated from post metadata (if a post has a density value then it has been scanned) - the value now is more in line with it's name.
* posts_with_issues is a new metric produced and holds the value that `posts_scanned` used to contain - this value is being kept under the new key as it may be useful in future.
* avg_issue_density_percentage now includes posts in the total that have `0` issues - so that the average reflects site average and not just the average density of posts that have issues.
* if avg_issues_per_page is < 1 then it will show `< 1` string rather than a number with a decimal.

Improvements to scan statistics:

* [`admin/class-scans-stats.php`](diffhunk://#diff-a60c9bd16cb44ab306c6325d9e5db0e79041354c491b655d3b616125dd1cdff1R159-R172): Added a direct database query to count the total unique meta values from the `postmeta` table where the `meta_key` is `_edac_issue_density`, providing the total number of posts that have been scanned. This replaces the previous method of counting distinct posts through `Issues_Query`.
* [`admin/class-scans-stats.php`](diffhunk://#diff-a60c9bd16cb44ab306c6325d9e5db0e79041354c491b655d3b616125dd1cdff1L262-R284): Updated the calculation of `avg_issue_density_percentage` to handle values less than 1, displaying them as `< 1` instead of rounding to zero. This change ensures more accurate representation of low issue densities.

## Checklist

## Summary by CodeRabbit

- **New Features**
	- Introduced a dismissible notice for improvements in statistics calculations on the admin welcome page.
	- Added AJAX functionality to handle dismissal of the welcome page notice.
	- Implemented a mechanism for counting distinct posts with issues for enhanced accuracy.
	- Enhanced event handling for the dismissal notice in the admin interface.
	- Added support for custom base links in URL generation.

- **Bug Fixes**
	- Improved accuracy of post scanning statistics calculation.
	- Enhanced calculation of average issue density percentage.
	- Refined data retrieval and formatting for scan statistics.

### DIFF
--- a/admin/class-admin-notices.php
+++ b/admin/class-admin-notices.php
@@ -61,16 +61,16 @@ class Admin_Notices {
 	/**
 	 * Notify users of the improvements to stats calculations.
 	 *
-	 * In version 1.20.0 we changed how posts_scanned was counted along with how other values like averages
+	 * In version 1.21.0 we changed how posts_scanned was counted along with how other values like averages
 	 * were calculated.
 	 *
-	 * @since 1.20.0
+	 * @since 1.21.0
 	 *
 	 * @return void
 	 */
 	public function welcome_page_post_count_change_notice() {
-		// Only show this notice if the version number is below 1.21.0.
-		if ( version_compare( EDAC_VERSION, '1.21.0', '>=' ) ) {
+		// Only show this notice if the version number is below 1.22.0.
+		if ( version_compare( EDAC_VERSION, '1.22.0', '>=' ) ) {
 			return;
 		}
 
--- a/admin/class-admin-notices.php
+++ b/admin/class-admin-notices.php
@@ -29,6 +29,9 @@ class Admin_Notices {
 
 		add_action( 'in_admin_header', [ $this, 'edac_remove_admin_notices' ], 1000 );
 		add_action( 'in_admin_header', [ $this, 'hook_notices' ], 1001 );
+		// register ajax actions.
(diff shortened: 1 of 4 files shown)


admin/class-admin-notices.php
```diff
@@ -54,6 +55,67 @@ public function hook_notices() {
...
+	 *
+	 * @since 1.20.0
+	 *
+	 * @return void
+	 */
+	public function welcome_page_post_count_change_notice() {
+		// Only show this notice if the version number is below 1.21.0.
+		if ( version_compare( EDAC_VERSION, '1.21.0', '>=' ) ) {
+			return;
+		}
```
pattonwebz, line 75:
```suggestion
	/**
	 * Notify users of the improvements to stats calculations.
	 *
	 * In version 1.21.0 we changed how posts_scanned was counted along with how other values like averages
	 * were calculated.
	 *
	 * @since 1.21.0
	 *
	 * @return void
	 */
	public function welcome_page_post_count_change_notice() {
		// Only show this notice if the version number is below 1.22.0.
		if ( version_compare( EDAC_VERSION, '1.22.0', '>=' ) ) {
			return;
		}
```

PR #863, opened 2025-03-12
Release v1.22.0

pattonwebz:
Changelog: 

William Patton (5):
- [x] Backport: Release/1.21.0
- [x] Bump the actions/cache to v4 to avoid fails due to v2 deprecated
- [x] New window warning migration
- [x] Add check if the theme "Tags" is an array - props @2ndkauboy 
- [x] Parse block content through the_content instead of do_block directly and use a new WP_Post object


SteveJonesDev (approved):
LGTM

PR #869, opened 2025-04-04
Add some extra code to shift focus in the skip link fix

pattonwebz:
This is to override bad smooth scroll implementations found on some sites which block the focus shifting so here a manual focus shift is added to combat that

## Checklist


SteveJonesDev (approved):
LGTM

PR #971, opened 2025-05-12
Fix metrics key from 'elementsCount' to 'elementCount' in density data

pattonwebz:
This pull request includes a small update to the `set_post_scan_results` method in the `class-rest-api.php` file. The change corrects the key name from `elementsCount` to `elementCount` when accessing the `$metrics` array.

This is the reason that density data wasn't being properly calculated.

## Checklist


SteveJonesDev (commented):
Screenshot 2025-05-12 at 2 16 21 PM

Looks like this fixes the issue.

PR #979, opened 2025-05-21
Make scan runnable from frontend loaded page

pattonwebz:
This PR adjusts the pageScanner bundle so that it can be invoked from the window object directly meaning it can be run from a loaded page on the frontend if a rescan is needed there.

## Checklist


SteveJonesDev (approved):
Tested locally and appears to work as expected and doesn't seem to interfere with previous iFrame scans. Nice cleanup of TODOS and console logs.

PR #1042, opened 2025-06-28
update: average issues per post calculation to handle zero issues case

SteveJonesDev:
This pull request introduces a small but meaningful change to improve the clarity of how average issues per post are displayed in the `admin/class-scans-stats.php` file. Specifically, it ensures that when there are no issues at all, the average is explicitly shown as `0`, and when the average is less than `1`, it is displayed as `"< 1"`.

Key change:

* Updated the logic in the `function ( $item )` to handle cases where there are no issues (`warnings` and `errors` both equal `0`) by explicitly setting `avg_issues_per_post` to `0`. This improves the accuracy and readability of the displayed data.


pattonwebz (approved):
I tested this and see why this change works. The old code was checking for exactly 0 integer but the return of `round` is a float so the check would need to be 0.0 or be loosely typed to get the same result.

PR #1063, opened 2025-07-08
Fix: Improve SQLite compatibility for WordPress Playground

pattonwebz:
NOTE: initially drafted by Jules

- Modified `edac_get_valid_table_name` to use SQLite-compatible query for checking table existence.
- Refactored `delete_cpt_posts` in `class-purge-post-data.php` to replace MySQL-specific multi-table DELETE syntax with a multi-step approach compatible with SQLite.
- Reviewed other SQL queries; they generally use standard SQL and $wpdb->prepare, requiring no changes for SQLite compatibility.

## Checklist

### DIFF
--- a/admin/class-purge-post-data.php
+++ b/admin/class-purge-post-data.php
@@ -81,16 +81,59 @@ class Purge_Post_Data {
 		}
 
 		global $wpdb;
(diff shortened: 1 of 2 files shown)


pattonwebz:
So many of our queries have changed or been updated since this PR was opened that when it comes time to do it most of the work here won't be the same.

PR #1105, opened 2025-07-24
Remove password protection features

SteveJonesDev:
## Summary
- drop password protection admin notice hooks
- remove password protected check from ajax callbacks
- stop exposing auth status in localized JS
- update editor script to always load scanner
- prune authorization details from site health
- clean up uninstall list and tests

## Testing
- `composer lint`
- `composer test` *(fails: Could not find WordPress test library)*

------
https://chatgpt.com/codex/tasks/task_e_68818e92d8a483289b677f416c9e350b

### DIFF
--- a/src/editorApp/index.js
+++ b/src/editorApp/index.js
@@ -7,14 +7,11 @@ window.addEventListener( 'DOMContentLoaded', () => {
 	// eslint-disable-next-line camelcase
 	const SCANNABLE_POST_TYPE = edac_editor_app.active;
 
-	if ( SCANNABLE_POST_TYPE ) {
-		// eslint-disable-next-line camelcase
-		if ( edac_editor_app.authOk === '1' ) {
-			setTimeout( function() {
-				initCheckPage();
-			}, 250 ); // Allow page load to fire before init, otherwise we'll have to wait for iframe to load.
-		}
-	}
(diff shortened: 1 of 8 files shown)


uninstall.php
```diff
@@ -32,8 +32,6 @@
 		'edac_frontend_highlighter_position',
 		'edac_delete_data',
 		'edac_review_notice',
-		'edac_authorization_password',
-		'edac_authorization_username',
```
pattonwebz, line 36: We should still keep these here so they get removed on plugin uninstall for anyone that had the plugin installed before we removed this feature, maybe with a commend noting they are legacy.

src/editorApp/index.js
```diff
@@ -7,14 +7,11 @@ window.addEventListener( 'DOMContentLoaded', () => {
 	// eslint-disable-next-line camelcase
 	const SCANNABLE_POST_TYPE = edac_editor_app.active;
 
-	if ( SCANNABLE_POST_TYPE ) {
-		// eslint-disable-next-line camelcase
-		if ( edac_editor_app.authOk === '1' ) {
```
pattonwebz, line 12: This is also defined in the pro plugin localize calls as well and should be removed from there too.

PR #1220, opened 2025-09-01
Use a specifically named ajax request for frontend highlighter

pattonwebz:
Release v1.31.0

## Checklist

Release v1.31.0

### DIFF
--- a/admin/class-ajax.php
+++ b/admin/class-ajax.php
@@ -602,6 +602,11 @@ class Ajax {
 
 		}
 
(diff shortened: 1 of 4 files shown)


SteveJonesDev (changes requested):
I think this looks good overall. Please test throughly as I'm unable to do so right now.

Nitpick:
In some place a single use variable is used and not in other places. Specifically for the WP_Error's. It probably should be consistent and in this case I'm not sure the variable helps readability.

PR #1247, opened 2025-10-01
updated: prefixes and text domain in PHPCS ruleset for accessibility checker

SteveJonesDev:
This pull request updates the coding standards configuration to match the plugin's actual naming and text domain conventions.

Configuration updates:

* Changed the global prefix in `.phpcs.xml.dist` from `my-plugin` to `edac` to ensure code uses the correct project-specific prefix.
* Updated the internationalization text domain in `.phpcs.xml.dist` from `my-plugin` to `accessibility-checker` for accurate localization checks.


pattonwebz (approved):
Approving the changes looks good, but consider removing the file instead. I don't use this file locally; it's not used in any workflows, and we also don't distribute it in the plugin distribution build. I'd be in favour of removing.

Issue #1151, opened 2025-07-31
Bug: Scanner attempts to scan unsaved new posts and gets 404 error


pattonwebz:
### Please give us a description of what happened.

When a new post is opened for the first time, the Accessibility Checker scanner tries to scan it immediately. However, since the post has not been saved yet—even as a draft—the scan request results in a 404 error page which probably gets scanned. The plugin should not attempt to scan posts until they have a valid post status.

### Please describe what you expected to happen and why.

I expected the Accessibility Checker to wait until the post has been saved (at least as a draft) and has a valid post status before attempting a scan. This would prevent unnecessary 404 errors and improve user experience.

### How can we reproduce this behavior?
1. Open the WordPress editor to create a new post.
2. Observe that the Accessibility Checker tries to scan immediately, before the post is saved.
3. The scan results in a 404 error.

PR #1278, opened 2025-11-25
Updated:  prevent initial scan on unsaved posts; fallback for classic editor

SteveJonesDev:
This pull request updates the logic for initializing the scan in `src/editorApp/checkPage.js` to ensure the scan only runs when the post has been saved. The main change is to prevent scans from running on auto-draft posts, improving both accuracy and resource usage.

Scan initialization improvements:

* Updated the `init` function to check the post status using `wp.data.select('core/editor').getEditedPostAttribute('status')`, and only run the scan if the post is not an auto-draft.
* Added a fallback for the classic editor, assuming the post is saved if on `post.php`, to maintain compatibility.

Fixes: https://linear.app/equalize-digital/issue/PRO-221/bug-scanner-attempts-to-scan-unsaved-new-posts-and-gets-404-error
Fixes: #1151

Gutenberg:

Classic:

## Summary by CodeRabbit

* **Bug Fixes**
  * Iframe injection is skipped for posts in the auto-draft state; initial scan is deferred until the post is saved or confirmed non-draft.
  * Ensures consistent iframe behavior when editor state is unavailable by using a fallback source for post status, avoiding unexpected injections.

### DIFF
--- a/src/editorApp/checkPage.js
+++ b/src/editorApp/checkPage.js
@@ -161,13 +161,10 @@ export const init = () => {
 	}
 
 	const editor = wp?.data?.select?.( 'core/editor' );
+	const postStatus = editor?.getEditedPostAttribute( 'status' ) ?? window?.edac_editor_app?.postStatus;
 
-	// In the block editor, don't run the initial scan on auto-drafts.
-	if ( editor ) {
-		const postStatus = editor.getEditedPostAttribute( 'status' );
-		if ( ! postStatus || postStatus === 'auto-draft' ) {
-			return;
-		}
+	if ( ! postStatus || postStatus === 'auto-draft' ) {
+		return;
 	}
 
 	// eslint-disable-next-line camelcase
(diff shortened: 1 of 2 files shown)


src/editorApp/checkPage.js
```diff
@@ -160,6 +160,17 @@ export const init = () => {
...
+
+	// In the block editor, don't run the initial scan on auto-drafts.
+	if ( editor ) {
+		const postStatus = editor.getEditedPostAttribute( 'status' );
+		if ( ! postStatus || postStatus === 'auto-draft' ) {
+			return;
+		}
+	}
+
+	// Fallback for classic editor or for saved posts in the block editor.
```
pattonwebz: This comment implies there's a fallback for the classic editor but there's no fallback below. Can we add one or remove the comment?

pattonwebz (approved):

Issue #1192, opened 2025-08-13
Front end highlighter doesn't account for global ignores


amberhinds:
Front end highlighter:

<img width="442" height="615" alt="Image" src="https://github.com/user-attachments/assets/7052bc0a-5687-433a-8b84-bcb799e5d03e" />

Editor for same post:

<img width="1047" height="694" alt="Image" src="https://github.com/user-attachments/assets/fa1133ef-1d4a-4f5d-bf37-e55612789e74" />

These items were all ignored globally and I expect that is the difference.

Should we even list the number of ignores on the front end highlighter given the issues can't be toggled through or seen on the front end once ignored? Perhaps the best solution is just to remove "0 ignored issues" from this view.

pattonwebz:
With the idea that we want to make most actions able to be done from the frontend highlighter (which may including managing ignores) I thing we shold try to fix the number rather than remove it.

amberhinds:
Once you have no open issues on a page it has a message that says something to the effect of "No issues detected." This already excludes ignores. It doesn't make sense to me to surface a number of ignores if you can't click through them.

I would prefer to remove the number from the front-end highlighter and figure out a way of surfacing and managing ignores on the front end as a pro feature in the alternate view we're going to build for pro.

SteveJonesDev:
This wasn't related to global ignores. When a global ignore is created, it actually updates the ignore column on individual issues as well. All logic works off the individual issues.

There was an actual bug on the comparison. I've fixed that and also set it to never show "0 ignored issues".

Ignored issues can actually be exposed in the frontend highlighter when you click 'View on page" of an ignored issue. Showing the count here is helpful for it to correctly add up to the total number of issues. It's kind of odd because it's only ever going to be "1 ignored issue".
<img width="953" height="647" alt="Image" src="https://github.com/user-attachments/assets/1705eade-0e32-4df9-9cec-a5150a939867" />

Ignore count removed when an ignored issue is not passed:
<img width="940" height="365" alt="Image" src="https://github.com/user-attachments/assets/43c55cca-0806-4a0e-9bce-2b7d40ca5572" />

PR #1290, opened 2025-12-02
Fixed: ignored issue count comparison and update ignored issues message logic.

SteveJonesDev:
This pull request addresses minor logic corrections in the `AccessibilityCheckerHighlight` class within `src/frontendHighlighterApp/index.js`. The changes improve the accuracy of issue counting and the display of ignored issues in the summary text.

Counting logic improvements:

* Updated the `countIgnored()` method to check for `'1'` (string) instead of `1` (number) when determining if an issue is ignored, ensuring correct counting behavior.

Summary display fix:

* Changed the summary text logic to only show ignored issues when `ignoredCount > 0`, preventing misleading output when there are zero ignored issues.

Fixes: #1192

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed ignored item counting behavior to ensure accurate tracking
  * Corrected summary panel text display to eliminate the "ignored issues" clause when no items are ignored

### DIFF
--- a/src/frontendHighlighterApp/index.js
+++ b/src/frontendHighlighterApp/index.js
@@ -926,7 +926,7 @@ class AccessibilityCheckerHighlight {
 	countIgnored() {
 		let count = 0;
 		for ( const issue of this.issues ) {
-			if ( issue.ignored === 1 ) {
+			if ( issue.ignored === '1' ) {
 				count++;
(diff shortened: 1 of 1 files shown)


pattonwebz (approved):
Nice find on the fix here!

PR #1291, opened 2025-12-02
Updated: passed test percentage logic to include scannable post types and handle cases with no posts scanned.

SteveJonesDev:
This pull request updates the scan statistics logic in `admin/class-scans-stats.php` to better handle scannable post types and calculation of passed percentage. The main improvements focus on ensuring that only relevant post types are considered and that the passed percentage is calculated more accurately.

**Scan statistics improvements:**

* The `Issues_Query` initialization now includes a `post_types` parameter set to the result of `Settings::get_scannable_post_types()`, ensuring that only scannable post types are queried.
* The calculation of `passed_percentage` now sets it to `'N/A'` by default and only computes a percentage if there are scannable post types and scanned posts, preventing misleading statistics when no relevant data is present.

Screenshot 2025-12-02 at 5 03 53 PM
Screenshot 2025-12-02 at 5 03 59 PM

Fixes: PRO-265

## Release Notes

* **Bug Fixes**
  * Scan statistics now correctly filter results to scannable post types only
  * Pass percentage displays 'N/A' when insufficient scan data is available, instead of potentially misleading values
  * Improved accuracy of pass percentage calculations


pattonwebz:
I updated this branch with the test changes from after 6.9 release. Test now pass again.

I also tested the code here and confirmed that the change does show N/A when there is no issues. I tested it by toggling off all my post types to get the issues cleared out.

Screenshot from 2025-12-16 13-34-49

PR #1304, opened 2025-12-11
Added: activation redirect to welcome page and add corresponding tests

SteveJonesDev:
This pull request introduces a new feature that redirects users to a welcome page after plugin activation, along with comprehensive tests to ensure the redirect only occurs under the correct conditions. The changes are organized to add the redirect logic, integrate it into the plugin's activation and admin flow, and thoroughly test all relevant scenarios.

**Activation Redirect Feature:**

* Added a new `Activation_Redirect` class in `admin/class-activation-redirect.php` that handles redirecting users to the welcome page after plugin activation, with safeguards to prevent unwanted redirects (e.g., during AJAX, REST, multisite, or bulk activation, and for users without the proper capability).
* Updated the plugin activation logic in `includes/activation.php` to set a transient (`edac_activation_redirect`) which triggers the redirect on the next admin page load.
* Hooked the new `Activation_Redirect` logic into the admin initialization process in `admin/class-admin.php` so it runs automatically for eligible users.

**Testing:**

* Added a comprehensive PHPUnit test suite in `tests/phpunit/Admin/ActivationRedirectTest.php` to verify the redirect logic, including tests for all major edge cases (e.g., AJAX, multisite, user permissions, bulk activation, and correct URL formation).

## Summary by CodeRabbit

* **New Features**
  * After activation, admins are redirected once to the Accessibility Checker welcome page; redirect is skipped for AJAX/REST, bulk/network activations, test environments, or users without sufficient permissions, and uses a short-lived activation flag.

* **Tests**
  * Added tests covering the activation redirect lifecycle, contextual guards, and the welcome-page URL.

### DIFF
--- a/tests/phpunit/Admin/ActivationRedirectTest.php
+++ b/tests/phpunit/Admin/ActivationRedirectTest.php
@@ -92,9 +92,14 @@ class ActivationRedirectTest extends WP_UnitTestCase {
 	}
 
 	/**
-	 * Test that transient is deleted after being checked.
+	 * Test that transient is deleted only when redirect happens.
+	 *
+	 * Note: This test expects a redirect exception to be thrown because
+	 * WP_TESTS_DOMAIN is defined in the test environment, preventing the
+	 * actual redirect. In a real scenario without WP_TESTS_DOMAIN, the
+	 * redirect would occur and the transient would be deleted.
 	 */
-	public function test_transient_is_deleted_after_check() {
+	public function test_transient_is_deleted_only_on_redirect() {
 		// Set the transient.
 		set_transient( 'edac_activation_redirect', true, 60 );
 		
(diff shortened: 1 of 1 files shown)


admin/class-activation-redirect.php
```diff
@@ -0,0 +1,109 @@
...
+	 * @return void
+	 */
+	public function maybe_redirect_to_welcome(): void {
+		// Check if the activation redirect transient exists.
+		if ( ! get_transient( 'edac_activation_redirect' ) ) {
+			return;
+		}
+
+		// Delete the transient immediately to prevent redirect loops.
+		delete_transient( 'edac_activation_redirect' );
```
pattonwebz: This would be better placed directly before the redirect instead of before all of these extra checks and early returns.

pattonwebz (changes requested):
I tested this, the code works. I get redirected when activating just the plugin. I don't get a redirect when activating several plugins at once.

I had just one comment that makes the redirect transient less likely to get cleared early.

pattonwebz (approved):

PR #1411, opened 2026-02-14
Guard SkipLinkFix against empty targets

SteveJonesDev:
## Summary
- initialize the skip-link targets array before normalization
- return early when normalized targets are empty so no invalid frontend fix payload is registered
- add a PHPUnit regression test for whitespace/comma-only target input

## Root Cause
`SkipLinkFix::run()` appended into `$targets` without initialization and then always captured it in a filter closure. When `edac_fix_add_skip_link_target_id` contained only delimiters/whitespace, `$targets` stayed undefined.

## Evidence Type
- failing-path unit test added in this run (`test_frontend_data_skips_empty_targets`)
- local verification gates passed (`lint:js`, `lint:php`, `test:php`)

## Risk Assessment
Low. Change is isolated to skip-link target normalization and only affects the empty-target edge case.

## Environment Limitations
None for this run.

## Release Notes

* **Bug Fixes**
  * Improved skip link validation to prevent processing when no valid targets are available.

* **Tests**
  * Added test coverage for empty skip link target handling.


pattonwebz (approved):
## Code Review Summary

**Verdict: Approve** — clean, correct, low-risk fix.

### What it does (correct)
1. Initializes `$targets = []` before the loop — fixes the undefined-variable bug when the target list normalizes to nothing
2. Early-returns when `$targets` is empty — prevents registering an invalid frontend-fix payload with zero valid targets
3. Adds a meaningful regression test with the exact failing input (whitespace/comma-only targets)

### Verified
- Root cause real: `$targets` was appended in a loop but never initialized; delimiter-only input left it undefined
- Isolated to skip-link target normalization — no ripple effects
- CI fully green (lint PHP 7.4–8.2, JS, integration, security, style)

### Minor (non-blocking, see inline)
- Could also assert the filter callback isn't registered, not just that data lacks the key

### Looks good
- Minimal diff (+20/−0)
- Correct use of `empty()` covering undefined + empty-after-filter
- Test exercises the exact failing path

---
*Reviewed by Hermes Agent*

includes/classes/Fixes/Fix/SkipLinkFix.php
```diff
@@ -143,6 +144,9 @@ public function run() {
 			}
 			$targets[] = '#' . $trimmed;
 		}
+		if ( empty( $targets ) ) {
```
pattonwebz, line 147: 💡 **Minor (non-blocking):** The test asserts `skip_link` is absent from the frontend data, which catches the regression. Even stronger would be asserting the `edac_filter_frontend_fixes_data` filter callback was never registered in the empty-target case (e.g. via `has_filter`), since that's the actual contract being guarded — the early return exists to avoid registering an invalid payload. Current coverage is sufficient; this would just lock the contract more tightly.

PR #1415, opened 2026-02-14
Fix frontend highlighter crash when issue ID is missing

SteveJonesDev:
## Summary
- guard showIssue so it exits gracefully when `this.issues` is uninitialized
- guard showIssue when an issue id is not found instead of dereferencing `undefined`
- add Jest coverage for both missing-data paths in showIssue

## Root cause
showIssue assumed the issue list was always initialized and that every requested id existed. In those edge cases it attempted to read issue.tooltip and issue.element from undefined, causing a runtime crash.

## Evidence type
- test: added regression coverage in `tests/jest/frontendHighlighterApp/showIssue.test.js` and confirmed full Jest + PHP test suites pass

## Risk assessment
- low risk: change is localized to showIssue fallback handling and does not alter the happy path for valid issues

## Environment limitations
- npm ci --ignore-scripts still reports an npm internal error (`Exit handler never called`) in this runner, but lint and test commands executed successfully with the current dependency state

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced error handling for accessibility issue detection to gracefully manage cases where issues are not found or unavailable on the page.

* **Tests**
  * Added comprehensive test coverage for edge cases in issue detection and handling.

### DIFF
--- a/src/frontendHighlighterApp/index.js
+++ b/src/frontendHighlighterApp/index.js
@@ -531,15 +531,27 @@ class AccessibilityCheckerHighlight {
 	 * @param {string} id - The ID of the element.
 	 */
 
-	showIssue = ( id ) => {
+	showIssue( id ) {
 		this.removeSelectedClasses();
 
 		if ( id === undefined ) {
 			return;
 		}
 
+		if ( ! Array.isArray( this.issues ) ) {
+			this.currentButtonIndex = -1;
+			this.currentIssueStatus = __( 'The element was not found on the page.', 'accessibility-checker' );
+			this.descriptionOpen( id );
+			return;
+		}
+
 		const issue = this.issues.find( ( i ) => i.id === id );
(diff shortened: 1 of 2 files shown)


SteveJonesDev:
Closing this PR — the crash fix has been superseded by changes already in `develop`.

**What was found:**

`develop` independently added a `if ( ! issue ) { return; }` early-return guard in `showIssue()`, which prevents the crash this branch was fixing. However, it does not include the `Array.isArray( this.issues )` guard or the user-facing "element not found" error message via `descriptionOpen()` that this branch introduced.

**Why closing:**

`develop` has diverged significantly from this branch (screen reader announcements, panel description refactor, UI updates, and more). Merging as-is would conflict heavily. The core crash is already avoided in `develop`. If the fuller UX behavior (showing an error message when an issue can't be found) is still desired, it should be re-applied as a fresh PR rebased on `develop`.

PR #1421, opened 2026-02-17
Fix user role intersection check to return false when permissions array is empty

pattonwebz:
This pull request refines the logic in the `edac_user_can_ignore()` function to more accurately determine if a user should be ignored based on their roles. The updated code now explicitly checks if there is any intersection between the user's roles and the ignored roles, and returns a boolean value. If there was an empty permissions array this would be empty array and thus be _truthy_.

User role checking improvement:

* Updated `edac_user_can_ignore()` in `includes/options-page.php` to return a boolean indicating whether the current user has any roles that should be ignored, instead of returning the raw intersection array or `false`.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Permission check for "ignore" now always returns a boolean, ensuring consistent role validation and avoiding accidental truthy non-boolean values.
* **Tests**
  * Added extensive unit tests covering admin bypass, multiple/partial role matches, empty/null configurations, users with no roles, case sensitivity, and strict boolean behavior.

### DIFF
--- a/includes/options-page.php
+++ b/includes/options-page.php
@@ -30,8 +30,13 @@ function edac_user_can_ignore() {
 	$user_roles        = ( isset( $user->roles ) ) ? $user->roles : [];
 	$ignore_user_roles = get_option( 'edacp_ignore_user_roles', [] );
 
-	$intersect = array_intersect( $user_roles, $ignore_user_roles );
-	return ! empty( $intersect );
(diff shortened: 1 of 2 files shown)


pattonwebz:
no longer needed after migrating to real capabilities instead of option stored grants

PR #1427, opened 2026-02-18
Fix warning in edac_filter_by_value for missing indexes

SteveJonesDev:
## Summary
- add a regression test for `edac_filter_by_value()` when an item is missing the requested index
- guard index access in `edac_filter_by_value()` so missing keys are skipped without raising warnings
- preserve existing return behavior by collecting only exact matches

## Root cause
`edac_filter_by_value()` read `$items[$key][$index]` directly. When an element lacked `$index`, PHP emitted an undefined array key warning.

## Evidence type
- test: added `FilterByValueTest::test_edac_filter_by_value_handles_missing_index`
- reproduced failure before fix as `RuntimeException: Undefined array key "status"` via warning-to-exception handling in test
- validated passing gates after fix: `npm run lint:js:fix`, `npm run lint:php:fix`, `npm run lint:js`, `npm run lint:php`, `npm run test:php`

## Risk assessment
Low risk. Change is limited to helper filtering logic and now skips non-array entries or rows missing the requested key. Existing behavior for matching rows is unchanged.

## Environment limitations
`git fetch` failed earlier in this run due to temporary DNS resolution for `github.com`; lint and test gates completed successfully.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved robustness of the filter functionality to properly handle missing array keys without producing warnings or notices.

### DIFF
--- a/includes/helper-functions.php
+++ b/includes/helper-functions.php
@@ -124,17 +124,21 @@ function edac_remove_element_with_value( $items, $key, $value ) {
  * @return array
  */
 function edac_filter_by_value( $items, $index, $value ) {
+	$newarray = [];
+
 	if ( is_array( $items ) && count( $items ) > 0 ) {
 		foreach ( array_keys( $items ) as $key ) {
-			$temp[ $key ] = $items[ $key ][ $index ];
+			if ( ! is_array( $items[ $key ] ) || ! array_key_exists( $index, $items[ $key ] ) ) {
+				continue;
+			}
 
-			if ( $temp[ $key ] === $value ) {
+			if ( $items[ $key ][ $index ] === $value ) {
 				$newarray[ $key ] = $items[ $key ];
 			}
 		}
 	}
 
-	if ( isset( $newarray ) && is_array( $newarray ) && count( $newarray ) ) {
+	if ( count( $newarray ) ) {
 		return array_values( $newarray );
 	}
 	return [];
--- a/tests/phpunit/helper-functions/FilterByValueTest.php
+++ b/tests/phpunit/helper-functions/FilterByValueTest.php
@@ -28,6 +28,49 @@ class FilterByValueTest extends WP_UnitTestCase {
 	}
 
 	/**
+	 * Ensures missing indexes do not trigger notices.
+	 */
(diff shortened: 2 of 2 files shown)


pattonwebz (approved):
## Code Review Summary

**Verdict: Approve** — solid defensive fix with an excellent regression test.

### What it does (correct)
1. Initializes `$newarray = []` up front — removes reliance on `isset()`
2. Skips non-array entries or rows missing the requested index (`!is_array(...) || !array_key_exists(...) → continue`)
3. Preserves strict `===` match behavior for rows that do have the index

### Verified
- Root cause real: direct `$items[$key][$index]` on a missing key → PHP 'Undefined array key' warning
- Behavior preserved for matching rows; missing-index rows now skipped (sensible)
- Regression test is excellent — converts PHP warnings to thrown exceptions, so the old bug fails loudly
- CI fully green (lint, integration, security; make-pot skip is expected)

### Minor (non-blocking, see inline)
- `count( $newarray )` check is now redundant (empty array → returns [] anyway) but harmless

### Looks good
- Minimal, isolated change to helper filtering
- Correct use of `array_key_exists` (vs isset, which would also treat null values as missing)
- Warning→exception test pattern is the right way to lock this in

---
*Reviewed by Hermes Agent*

includes/helper-functions.php
```diff
@@ -124,17 +124,21 @@ function edac_remove_element_with_value( $items, $key, $value ) {
...
+			if ( ! is_array( $items[ $key ] ) || ! array_key_exists( $index, $items[ $key ] ) ) {
+				continue;
+			}
 
-			if ( $temp[ $key ] === $value ) {
+			if ( $items[ $key ][ $index ] === $value ) {
 				$newarray[ $key ] = $items[ $key ];
 			}
 		}
 	}
```
pattonwebz, line 139: 💡 **Minor (non-blocking):** The trailing `if ( count( $newarray ) )` is now redundant — since `$newarray` is always initialized to `[]` and the function returns `array_values($newarray)` or `[]` at the end, an empty array naturally falls through to returning `[]`. Keeping it is harmless (preserves the exact return shape), so no change required — just noting it could be simplified to a plain `return array_values( $newarray );`.

PR #1534, opened 2026-03-09
Add scannable status option

pattonwebz:
This pull request introduces support for configuring which post statuses are scanned by the accessibility checker, allowing users to select statuses such as 'publish', 'draft', or 'private' via the settings UI. It also adds a mechanism to purge scan data for posts of a removed status and ensures robust handling of status selection, including filter overrides and sanitization. Comprehensive tests have been added to verify these behaviors.

**Post Status Configuration Enhancements**

* Added a new option (`edac_post_statuses`) to allow users to select which post statuses should be scanned, including UI, registration, and default values. [[1]](diffhunk://#diff-9f51a5a008e12b8352a5578ecf0df8ab43937bb5e9f1c9ec49aeb769b5fbba2cL27-R30) [[2]](diffhunk://#diff-92fd73fa85986ecef05555786d999392e201dd161e2c0e63b1e0bd3a139ce954R165-R173) [[3]](diffhunk://#diff-92fd73fa85986ecef05555786d999392e201dd161e2c0e63b1e0bd3a139ce954R293) [[4]](diffhunk://#diff-92fd73fa85986ecef05555786d999392e201dd161e2c0e63b1e0bd3a139ce954R690-R800) [[5]](diffhunk://#diff-b825f493538a13f917f49c96a5f34aefab1d1e5f5143b191b12ba07c72718678R21-R35)
* Implemented a settings UI (`edac_post_statuses_cb`) for post status selection, with filter detection to disable the UI when overridden by `edac_scannable_post_statuses`.

**Data Purging and Sanitization**

* Added `Purge_Post_Data::delete_status_posts` method to purge accessibility checker data for posts of a specific status, and integrated it into the sanitization logic to remove scan data when a status is deselected. [[1]](diffhunk://#diff-0a0dc9e705455aa9483cdc99ada782e0224a1d54c231c57dcc8bda9b1aa06744R68-R109) [[2]](diffhunk://#diff-92fd73fa85986ecef05555786d999392e201dd161e2c0e63b1e0bd3a139ce954R690-R800)
* Enhanced sanitization (`edac_sanitize_post_statuses`) to ensure only allowed statuses are saved, preserve stored values if filter is active, and clear cached stats and scan data when statuses change.

**Testing Improvements**

* Added PHPUnit tests for the new purge method and sanitization logic, covering data removal, filter overrides, and input validation. [[1]](diffhunk://#diff-f79c21ba2757a1cca7327421100f5d10ca34b45b5afd52366465de1aaca4fde4R191-R279) [[2]](diffhunk://#diff-39bfe43781656cd9e091399cfb56a1c969fe366113b33fbc4f29611f0aa5e5ceR1-R76)

**Miscellaneous**

* Registered the new option for deletion on uninstall.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added "Post Statuses To Be Checked" admin setting with sanitization and filter support.
  * New status-based purge: removing a status can automatically purge related stored scan data.

* **Chores**
  * Installer initializes the post-statuses setting; uninstaller removes it when site data deletion is enabled.

* **Tests**
  * Added unit tests for status selection, sanitization, persistence, and the status-based purge behavior.


tests/phpunit/includes/classes/OptionsPagePostStatusesTest.php
```diff
@@ -0,0 +1,140 @@
...
+ */
+class OptionsPagePostStatusesTest extends WP_UnitTestCase {
+
+	/**
+	 * Create the custom table that delete_status_posts() needs.
+	 *
+	 * @return void
+	 */
+	public function set_up(): void {
+		parent::set_up();
```
SteveJonesDev, line 19: `setUp` uses camelCase; `set_up` uses snake_case inconsistently. I searched the plugin and it looks like we aren't consistent with this. We may want to create a follow-up PR to address that.

includes/options-page.php
```diff
@@ -689,6 +699,136 @@ function edac_post_types_cb() {
...
+
+	$stored_statuses = get_option( 'edac_post_statuses', $all_statuses );
+	if ( ! is_array( $stored_statuses ) || empty( $stored_statuses ) ) {
+		$stored_statuses = $all_statuses;
+	}
+
+	if ( is_array( $selected_statuses ) ) {
+		$selected_statuses = array_values( array_intersect( $selected_statuses, $all_statuses ) );
+	} else {
+		$selected_statuses = $all_statuses;
```
SteveJonesDev, line 806: If $selected_statuses is not an array (e.g., a stale string from the DB), the function silently resets to all statuses. This could be surprising. At minimum, it's worth a comment explaining that this is an intentional fallback.

Possibly related: If a user deselects all post statuses and clicks Save Changes on refresh, all statuses are selected again. This is not how other checkbox settings function.

admin/class-settings.php
```diff
@@ -18,17 +18,21 @@ class Settings {
 	 * @var array
 	 */
 	public static function get_scannable_post_statuses() {
+
+		$statuses = Helpers::get_option_as_array( 'edac_post_statuses' );
+
+		if ( empty( $statuses ) ) {
+			$statuses = [ 'publish', 'future', 'draft', 'pending', 'private' ];
```
SteveJonesDev, line 25: Default statuses are hardcoded in multiple places. 

`[ 'publish', 'future', 'draft', 'pending', 'private' ]` appears in:

edac_post_statuses_cb()
edac_sanitize_post_statuses()
activation.php
class-settings.php

This is a maintenance concern. A constant or a small helper function (e.g., edac_get_default_post_statuses()) would centralize this.

includes/options-page.php
```diff
@@ -689,6 +699,136 @@ function edac_post_types_cb() {
...
+	];
+
+	$fieldset_attrs = '';
+	if ( $is_filter_active ) {
+		$fieldset_attrs = 'aria-disabled="true"';
+	} elseif ( ! $is_pro ) {
+		$fieldset_attrs = 'class="edac-setting--upsell"';
+	}
+	?>
+	<fieldset <?php echo $fieldset_attrs; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped ?>>
```
SteveJonesDev, line 742: `$fieldset_attrs` is built from two known static strings, so it's not a security risk in practice, but it's a bit of a code smell. Consider using `wp_kses` with an allowed attribute list or splitting into discrete `echo esc_attr(...)` calls per attribute.

admin/class-purge-post-data.php
```diff
@@ -65,6 +65,48 @@ public static function delete_post_meta( int $post_id ) {
...
+	 *
+	 * @since 1.38.0
+	 *
+	 * @param string $post_status Post status slug (e.g. 'draft', 'private').
+	 *
+	 * @return bool|int|\mysqli_result|void
+	 */
+	public static function delete_status_posts( string $post_status ) {
+
+		if ( ! $post_status ) {
```
SteveJonesDev, line 83: This guard might be a bit weak; should we explicitly return `false`?

If `( ! $post_status )` returns early for an empty string, but the parameter is typed `string`, so `null` can't be passed. More importantly, it silently returns `void` while the function's declared return type suggests a result is expected. Consider returning `false` explicitly or throwing for invalid input to be consistent.

admin/class-purge-post-data.php
```diff
@@ -65,6 +65,48 @@ public static function delete_post_meta( int $post_id ) {
...
+		 * @since 1.38.0
+		 *
+		 * @param string $post_status Post status.
+		 */
+		do_action( 'edac_before_delete_status_posts', $post_status );
+
+		// phpcs:ignore WordPress.DB.DirectDatabaseQuery.DirectQuery, WordPress.DB.DirectDatabaseQuery.NoCaching -- Safe variable used for table name, caching not required for one time operation.
+		return $wpdb->query(
+			$wpdb->prepare(
+				"DELETE T1,T2 from $wpdb->postmeta as T1 JOIN %i as T2 ON T1.post_id = T2.postid JOIN $wpdb->posts as P ON P.ID = T1.post_id WHERE T1.meta_key like %s and T2.siteid=%d and P.post_status=%s",
```
SteveJonesDev, line 101: I don't fully understand the AI reviews regarding the SQL query, but can we confirm it's not a real issue?

From Claude:
The DELETE T1,T2 joins postmeta → custom table → posts, but the WHERE clause filters T2.siteid and P.post_status — it does not filter the custom table (T2) by any column related to what's in T1. The join on T1.post_id = T2.postid means it deletes rows from T2 whose postid matches posts with the given status, which is correct. But notice the query also deletes from T1 (postmeta) via the JOIN — this means it will delete _edac* meta for those posts. That part is working as intended, but worth confirming the join condition T1.post_id = T2.postid is reliable when postmeta rows exist for posts not in the custom table.

More importantly: The query has no JOIN $wpdb->posts equivalent for the T2 rows — T2 rows are only deleted if they match a T1 (postmeta) row. If a post has a custom table row but no _edac* postmeta, its custom table row won't be purged.

```php
// Current (problematic if postmeta is absent):
"DELETE T1,T2 from $wpdb->postmeta as T1
  JOIN %i as T2 ON T1.post_id = T2.postid
  JOIN $wpdb->posts as P ON P.ID = T1.post_id
  WHERE T1.meta_key like %s and T2.siteid=%d and P.post_status=%s"
Consider two separate queries (matching the pattern of delete_post_meta):
```
```php
// 1. Delete custom table rows for posts of this status
"DELETE T FROM %i as T JOIN $wpdb->posts as P ON P.ID = T.postid
 WHERE T.siteid = %d AND P.post_status = %s"

// 2. Delete postmeta for posts of this status
"DELETE M FROM $wpdb->postmeta as M JOIN $wpdb->posts as P ON P.ID = M.post_id
 WHERE M.meta_key LIKE %s AND P.post_status = %s"
```

SteveJonesDev (changes requested):

PR #1716, opened 2026-05-26
Fix: skip img empty alt warning when image is inside a button with an accessible name

### LINKED ISSUE #1700: Enhance Image empty alt rule
In an ideal world, we would not flag a needs review issue on an image like this that has empty alt but that is in a button with an aria-lable:

```
<button aria-label="Close" class="cky-banner-btn-close" data-cky-tag="close-button"><img alt="" src="https://cdn-cookieyes.com/assets/images/close.svg" class="edac-highlight-element-selected edac-highlight-element-selected-min-width"></button>
```

Image

SteveJonesDev:
## Summary

- Adds an exception to the `img_alt_empty` check so images with `alt=""` inside a `<button>` (or `[role="button"]`) that has a non-empty `aria-label` or `title` are no longer flagged as violations.
- Adds 4 new Jest test cases covering the button-with-aria-label, button-with-title, role="button"-with-aria-label, and button-with-no-accessible-name scenarios.

## Why

Closes #1700. An image with `alt=""` inside a labelled button is intentionally decorative — the button's `aria-label` already provides the accessible name for the interactive element. Flagging this as a "needs review" issue is a false positive.

This mirrors the existing exception already in place for `<a>` elements: if an anchor has `aria-label`, `title`, or sufficient text content, the empty-alt image inside it is skipped. The same logic applies to buttons.

**Example that was incorrectly flagged:**
```html
<button aria-label="Close" class="cky-banner-btn-close">
  <img alt="" src="close.svg">
</button>
```

## What reviewers should know

- The fix is in `src/pageScanner/checks/img-alt-empty-check.js` inside the existing `isInsideValidCaption` helper (which already handles anchors, figures, and WP captions).
- A button **without** an accessible name still triggers the warning — only labelled buttons are exempted.
- All 18 Jest tests pass.

Fixes PRO-811

## Summary by CodeRabbit

* **New Features**
  * Images inside button-like controls that have an accessible name (aria-label, aria-labelledby, title, or visible text) are now treated as having sufficient captioning for accessibility checks, reducing false empty-alt warnings.

* **Tests**
  * Added test coverage for image validation in button contexts, covering cases with and without accessible names.

[Review Change Stack](https://app.coderabbit.ai/change-stack/equalizedigital/accessibility-checker/pull/1716?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/pageScanner/checks/img-alt-empty-check.js
+++ b/src/pageScanner/checks/img-alt-empty-check.js
@@ -65,12 +65,18 @@ function isInsideValidCaption( node ) {
 	}
 
 	// Check if inside button with valid accessible name
-	const button = node.closest( 'button, [role="button"]' );
+	const button = node.closest( 'button, [role~="button"]' );
 	if ( button ) {
 		if ( button.hasAttribute( 'aria-label' ) && button.getAttribute( 'aria-label' ).trim() !== '' ) {
 			return true;
 		}
+		if ( button.hasAttribute( 'aria-labelledby' ) && button.getAttribute( 'aria-labelledby' ).trim() !== '' ) {
+			return true;
+		}
 		if ( button.hasAttribute( 'title' ) && button.getAttribute( 'title' ).trim() !== '' ) {
+			return true;
+		}
+		if ( button.textContent.trim() !== '' ) {
 			return true;
 		}
 	}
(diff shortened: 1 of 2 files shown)


pattonwebz (approved):
LGTM 🚢

PR #1720, opened 2026-05-27
Feature: add Equalize Digital tab to WP 7.0+ Plugins screen

SteveJonesDev:
## What changed

Adds a new `Plugin_List_Tab` class (`admin/class-plugin-list-tab.php`) and wires it into `Admin::init()`.

The class hooks into two filters introduced in WordPress 7.0:

- **`plugins_list`** — appends an `equalize-digital` status group containing every installed plugin whose `AuthorURI` contains `equalizedigital.com`. The tab only appears when 2 or more such plugins are found (matching the convention used by Yoast and others).
- **`plugins_list_status_text`** — returns the "Equalize Digital" label for that tab.

Both hooks are version-gated behind `version_compare( $wp_version, '7.0-alpha0', '<' )` so the behaviour is a no-op on older WordPress installs.

## Why

WordPress 7.0 added `plugins_list` and `plugins_list_status_text` to let plugin authors surface a branded tab on the Plugins screen for their suite of plugins. This gives Equalize Digital users a convenient way to filter and view all Equalize Digital plugins in one place.

## Reviewer notes

- Detection is done via `AuthorURI` containing `equalizedigital.com` — reliable across both the free and Pro plugins without hardcoding slugs.
- Minimum threshold is 2 plugins (`MINIMUM_FOR_TAB`), so the tab won't appear for users who only have one Equalize Digital plugin installed.
- No changes to existing behaviour on WordPress < 7.0.

## Test plan

- [ ] Install two or more Equalize Digital plugins on a WordPress 7.0+ site and verify the "Equalize Digital" tab appears on the Plugins screen
- [ ] Verify the tab correctly filters to only Equalize Digital plugins
- [ ] Confirm no tab appears when only one Equalize Digital plugin is installed
- [ ] Confirm no regressions on WordPress < 7.0 (hooks should not be registered)

## Summary by CodeRabbit

* **New Features**
  * Adds a custom "Equalize Digital" tab on the WordPress Plugins screen (visible on WordPress 7.0+ when three or more Equalize Digital plugins are installed).
  * Shows a plural-aware, accessibility-friendly status label that reflects the number of grouped plugins and reduces noise by requiring a higher threshold before grouping.

[Review Change Stack](https://app.coderabbit.ai/change-stack/equalizedigital/accessibility-checker/pull/1720?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/admin/class-plugin-list-tab.php
+++ b/admin/class-plugin-list-tab.php
@@ -5,7 +5,7 @@
  * @package Accessibility_Checker
  */
 
-namespace EDAC\Admin;
+namespace EqualizeDigital\AccessibilityChecker\Admin;
 
 if ( ! defined( 'ABSPATH' ) ) {
 	exit;
--- a/admin/class-plugin-list-tab.php
+++ b/admin/class-plugin-list-tab.php
@@ -90,18 +90,18 @@ class Plugin_List_Tab {
 	 *
 	 * @since 1.43.0
 	 *
-	 * @param string $text  The current status text.
-	 * @param int    $count The number of plugins with this status.
-	 * @param string $type  The status type slug.
+	 * @param string $text   The current status text.
+	 * @param int    $_count Unused — "Equalize Digital" does not inflect by count.
+	 * @param string $type   The status type slug.
 	 *
 	 * @return string The status text.
 	 */
-	public function get_status_text( $text, $count, $type ) {
+	public function get_status_text( $text, $_count, $type ) {
 		if ( self::STATUS_SLUG !== $type ) {
 			return $text;
 		}
 
-		return \_nx( 'Equalize Digital', 'Equalize Digital', $count, 'plugin status', 'accessibility-checker' );
+		return \_\_( 'Equalize Digital', 'accessibility-checker' );
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


admin/class-plugin-list-tab.php
```diff
@@ -0,0 +1,123 @@
...
+	 * @param string $type  The status type slug.
+	 *
+	 * @return string The status text.
+	 */
+	public function get_status_text( $text, $count, $type ) {
+		if ( self::STATUS_SLUG !== $type ) {
+			return $text;
+		}
+
+		return \_nx( 'Equalize Digital', 'Equalize Digital', $count, 'plugin status', 'accessibility-checker' );
```
pattonwebz: Not sure why Gemini suggested using `_nx` here because this can't every actually be reached in the singluar form.

Seems like a waste to translate it to the same no matter if singular or plural so I would just translate it with __ directly and ignore the Gemini review.

pattonwebz (changes requested):
Just one comment about unneeded _nx use. Otherwise seems good to me.

SteveJonesDev:
Holding on this PR to see if this Trac ticket makes it into core. https://core.trac.wordpress.org/ticket/65359

PR #1723, opened 2026-05-28
🧪 POC: Per-rule exclusions config for JS scanner checks

### LINKED ISSUE #567: Additional core block exclusions on aria-hidden
## Is your feature request related to a problem? Please describe.

There are some correct uses of aria-hidden related to core blocks that it would be good for us to exclude from flagging an aria-hidden warning because we know that they are correct.

## Describe the solution you'd like
Screenshot 2024-03-30 at 9 38 13 PM

Here are the code snippets:
`<span aria-hidden="true" class="wp-block-cover__background has-background-dim"></span>`

`<span class="wp-block-post-featured-image__overlay has-background-dim has-background-dim-10 has-contrast-2-background-color" style="border-radius:5px" aria-hidden="true"></span>`

## Why do you think this feature is something we should consider for this plugin?

We should avoid as many false positives as we can and make our rules the best ruleset for WordPress. 🙂

SteveJonesDev:
## Summary

- Adds an `exclusions` key to PHP rule configs, allowing WordPress-specific CSS selectors to be declared at the rule level rather than hardcoded in generic JS check logic
- Threads those exclusions through the existing `scanOptions` pipeline (PHP → `wp_localize_script` → iframe `window.scanOptions`) so the JS check can apply them via `node.matches()`
- Removes the hardcoded `wp-block-spacer` check from `aria-hidden-valid-usage.js` and replaces it with the new config-driven approach; adds the two new WP core block overlay patterns from issue #567 at the same time

## Motivation

The JS scanner rules are used as a standalone package outside of WordPress. Hardcoding WordPress class names (like `wp-block-spacer`) directly in the JS checks pollutes generic accessibility logic with framework-specific knowledge. This approach keeps the JS checks clean and moves all WordPress-specific exclusion knowledge into the PHP rule config where it belongs.

## How it works

**PHP rule config** (`AriaHiddenRule.php`):
```php
'exclusions' => [
    '.wp-block-spacer',
    '.wp-block-cover__background.has-background-dim',
    '.wp-block-post-featured-image__overlay.has-background-dim',
],
```

**Collected at enqueue time** (`class-enqueue-admin.php`) and passed as `ruleExclusions` in `edac_editor_app`.

**Forwarded to iframe** (`checkPage.js`) via `window.scanOptions.ruleExclusions`.

**Applied in JS check** (`aria-hidden-valid-usage.js`) using `node.matches(selector)` — no WordPress knowledge in the check itself.

## Test plan

- [ ] Verify cover block background dim spans no longer trigger aria-hidden warnings
- [ ] Verify post featured image overlay spans no longer trigger aria-hidden warnings  
- [ ] Verify legitimate aria-hidden misuse still flags correctly
- [ ] Verify `wp-block-spacer` elements still pass (now via config rather than hardcoded)
- [ ] Confirm the JS package used outside WordPress still functions with an empty exclusions array

Closes #567

Fixes PRO-475

## Summary by CodeRabbit

* **New Features**
  * Added configurable exclusions for aria-hidden validation, enabling administrators to exclude specific elements and block types from accessibility checks, providing more tailored validation results.

[Review Change Stack](https://app.coderabbit.ai/change-stack/equalizedigital/accessibility-checker/pull/1723?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/admin/class-enqueue-admin.php
+++ b/admin/class-enqueue-admin.php
@@ -133,18 +133,19 @@ class Enqueue_Admin {
 					'edac-editor-app',
 					'edac_editor_app',
 					[
-						'postID'       => $post_id,
-						'edacUrl'      => esc_url_raw( get_site_url() ),
-						'edacApiUrl'   => esc_url_raw( rest_url() . 'accessibility-checker/v1' ),
-						'baseurl'      => plugin_dir_url( __DIR__ ),
-						'active'       => $active,
-						'pro'          => $pro,
-						'debug'        => $debug,
-						'scanUrl'      => $scan_url,
-						'maxAltLength' => max( 1, absint( apply_filters( 'edac_max_alt_length', 300 ) ) ),
-						'version'      => EDAC_VERSION,
-						'postStatus'   => get_post_status( $post_id ),
-						'restNonce'    => wp_create_nonce( 'wp_rest' ),
+						'postID'         => $post_id,
+						'edacUrl'        => esc_url_raw( get_site_url() ),
+						'edacApiUrl'     => esc_url_raw( rest_url() . 'accessibility-checker/v1' ),
+						'baseurl'        => plugin_dir_url( __DIR__ ),
+						'active'         => $active,
+						'pro'            => $pro,
+						'debug'          => $debug,
+						'scanUrl'        => $scan_url,
+						'maxAltLength'   => max( 1, absint( apply_filters( 'edac_max_alt_length', 300 ) ) ),
+						'ruleExclusions' => self::get_rule_exclusions(),
+						'version'        => EDAC_VERSION,
+						'postStatus'     => get_post_status( $post_id ),
+						'restNonce'      => wp_create_nonce( 'wp_rest' ),
 					]
 				);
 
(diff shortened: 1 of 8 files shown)


SteveJonesDev:
Addressed all inline review feedback in follow-up commits:

**`aria-hidden-valid-usage.js` (comments by @gemini-code-assist and @coderabbitai):** Added `Array.isArray()` guard and wrapped `node.matches()` in `try/catch` so malformed selectors can't crash the scan.

**`checkPage.js` (@gemini-code-assist):** Now spreads existing `scanOptions` before assigning new keys — nothing previously set gets overwritten.

**`class-enqueue-admin.php` (@gemini-code-assist):** Added `is_array()` + non-empty `$rule['slug']` checks in `get_rule_exclusions()` as defensive guards.

**Frontend scan regression (`config/rules.js`, @chatgpt-codex-connector):** `ruleExclusions` is now passed via `edacFrontendHighlighterApp` in `class-enqueue-frontend.php` and wired into `window.scanOptions` in `frontendHighlighterApp/index.js` before the scan runs — frontend rescans now get the same exclusions as editor iframe scans.

**Jest failure:** The `ariaHiddenValid.test.js` `beforeAll` wasn't passing check options, so the `wp-block-spacer` test broke when the hardcoded class check was removed. Updated the test to configure axe with the WP exclusions via check `options`, mirroring what `config/rules.js` does at runtime. Also added test cases for the two new overlay selectors from issue #567.

PR #1745, opened 2026-06-16
PRO-959: Add edac_get_landmark_types() as single source of truth for landmark detection

SteveJonesDev:
## Summary
- The scanner's landmark detection rules (`LANDMARK_TAGS`, `LANDMARK_ROLES`, `CONDITIONAL_LANDMARK_TAGS`, `CONDITIONAL_LANDMARK_ROLES` in `src/pageScanner/index.js`) were hardcoded inline, with the pro plugin maintaining its own separate, incomplete, out-of-sync copy for its Landmark filter dropdown.
- Added `edac_get_landmark_types()` in `includes/helper-functions.php` as the canonical PHP source, plus `edac_get_landmark_filter_options()` which flattens it into `{ value, label }` pairs for filter UIs.
- `pageScanner/index.js` now reads its detection constants from `window.scanOptions` (matching the existing `maxAltLength` override pattern) with the same hardcoded values kept only as a fallback for contexts that don't localize `scanOptions` (e.g. standalone test runs).
- Wired `edac_get_landmark_types()` into the editor scan's `wp_localize_script` payload (`edac_editor_app.landmarkTypes`) and into `checkPage.js`'s iframe `scanOptions` injection, alongside the existing `maxAltLength` override.

Linear: https://linear.app/equalize-digital/issue/PRO-959/landmark-filter-options-replace-hardcoded-lists-with-a-single-source

Companion PR in accessibility-checker-pro consumes `edac_get_landmark_filter_options()` for the Issues Explorer's Landmark filter.

## Test plan
- [ ] Run an editor "Check" scan on a post with `<nav>`, `<aside>`, `role="navigation"`, and `role="complementary"` elements; confirm the saved `landmark` values match the expected tag/role strings (unchanged behavior — same hardcoded defaults, now sourced via localization)
- [ ] Confirm a scan with no `wp_localize_script` data present (e.g. direct script load without the editor flow) still works via the fallback constants

## Summary by CodeRabbit

* **New Features**
  * Landmark detection now supports custom configuration options, allowing dynamic adjustments to landmark tag and role matching for page scanning.
  * Landmark type definitions and filter options are now available throughout the editor and admin interface, enabling consistent landmark handling across all scanning operations.


SteveJonesDev:
Closing — the proposed single-source-of-truth approach needs rework (edac_get_landmark_filter_options shouldn't live in the free plugin, it's pro-only UI formatting). Leaving PRO-959 open to redo.

PR #1791, opened 2026-06-23
Normalize WordPress.org deploy version from release tag

pattonwebz:
## Summary
- add a release-version normalization step in the dotorg deploy workflow
- strip leading and trailing 'v' characters from the release tag before passing VERSION

## Why
A release tag containing a v prefix/suffix produced an incorrect WordPress.org SVN tag.

## Summary by CodeRabbit

* **Chores**
  * Improved release deployment workflow to normalize version formatting, ensuring consistent and reliable handling of version identifiers during automated releases.

### DIFF
--- a/.github/workflows/deploy-on-release-to-dot-org.yml
+++ b/.github/workflows/deploy-on-release-to-dot-org.yml
@@ -50,6 +50,17 @@ jobs:
           npm run dist:dotorg
           echo "::set-output name=zip-path::./dist/${{ github.event.repository.name }}/${{ github.event.repository.name }}.zip"
 
+      - name: Normalize release version
+        id: release-version
+        run: |
+          RAW_VERSION="${{ github.event.release.tag_name }}"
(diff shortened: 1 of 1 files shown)


pattonwebz:
Closing as superseded: develop already contains d9e4de85 ("fix: strip leading v from release tag before using as SVN version"), which landed after this branch was cut and does the same core normalization — that competing edit at the same spot in the workflow is also why this PR shows a merge conflict.

The two extras here don't survive scrutiny:
- The GITHUB_REF_NAME fallback for empty tag names (real on workflow_dispatch runs) would pass a branch name like "main" as the deploy VERSION, creating a bogus SVN tag — worse than the empty VERSION the 10up action handles by deriving the version from the plugin file.
- The trailing-v strip has no real-world input.

If we want to harden the manual-dispatch path, that should be a small separate PR that derives the version from the plugin header (or fails loudly) rather than falling back to the ref name.

PR #1817, opened 2026-07-02
PRO-1168: fix: don't flag .ogg audio as video content in video_present rule

### LINKED ISSUE #1816: Audio block detected as Vídeo block resulting in "Video is present" warning
- [X] I've read and understood the [contribution guidelines](<https://github.com/equalizedigital/accessibility-checker/blob/develop/.github/CONTRIBUTING.md>).
- [X] I've searched for any related issues and avoided creating a duplicate issue.

### Please give us a description of what happened.

I'm testing a post with every block I support, and the **Audio** block appears to trigger a **Video** present warning.

### Please describe what you expected to happen and why.

I expected a **Audio present** or **Media present** warning.

### How can we reproduce this behavior?

1. Insert an audio block with an MP3 file
2. Save post
3. You should get a "Video present" warning.

### Technical info

* WordPress version: 7.0
* Accessibility Checker version: 1.45.0
* If relevant, which editor is affected (or editors):

- [ ] Classic Editor
- [X] Gutenberg
- [ ] Classic Editor plugin
- [ ] Page Builder (please specify):

* Which browser is affected (or browsers):

- [ ] Internet Explorer
- [ ] Edge
- [ ] Chrome
- [X] Firefox
- [ ] Safari
- [ ] Brave

* Relevant plugins in case of a bug:
* Tested with theme: Blocksy

Image

pattonwebz:
## Summary
- The Gutenberg Audio block's Ogg Vorbis sample file (`.ogg`) was being detected as video content by the `video_present` rule, because `.ogg` is also a valid Ogg Theora video extension.
- `.ogg` stays in `videoExtensions` (Ogg can still carry Theora video), but a `.ogg` match is now only counted as video when it isn't attached to an `<audio>` element or one of its `<source>` children.

## Test plan
- [x] Added regression tests reproducing the bug (plain `<audio src="....ogg">`, the real `wp-block-audio` figure markup, and a `<source>` inside `<audio>`) — confirmed these failed before the fix.
- [x] Added tests locking in that `.ogg` still triggers the rule outside an audio container (native `<video>`, standalone `<source>`).
- [x] `npx jest --config=./tests/jest/jest.config.js` — 847/847 passing, no regressions.

Fixes #1816.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved video detection to avoid falsely flagging `.ogg` sources that appear within audio contexts (such as audio elements and audio block structures).
  * Continued to flag `.ogg` when it’s used in actual video-related scenarios, including standalone video sources and `<video>` elements.
* **Tests**
  * Expanded rule coverage with additional `.ogg` cases to confirm both violations and non-violations across relevant HTML structures.

### DIFF
--- a/src/pageScanner/checks/is-video-detected.js
+++ b/src/pageScanner/checks/is-video-detected.js
@@ -33,14 +33,27 @@ export default {
 			}
 		}
 
+		// .ogg is the Ogg Vorbis audio extension, but the Ogg container can also carry
+		// Theora video, so it stays in videoExtensions. To avoid flagging ordinary Ogg
+		// Vorbis audio (e.g. the Gutenberg Audio block), only count a .ogg match as video
+		// when it isn't attached to an <audio> element or one of its <source> children.
+		const parentTag = node.parentNode ? node.parentNode.nodeName.toLowerCase() : '';
+		const isInAudioContainer = tag === 'audio' || ( tag === 'source' && parentTag === 'audio' );
+
 		const matchesExtension = videoExtensions.some( ( ext ) => {
 			const srcLower = src.toLowerCase();
 			const dataLower = data.toLowerCase();
 			// Check if the extension is at the end of the string or followed by a query parameter
-			return (
+			const matches = (
 				( srcLower.endsWith( ext ) || srcLower.includes( ext + '?' ) ) ||
 				( dataLower.endsWith( ext ) || dataLower.includes( ext + '?' ) )
 			);
+
+			if ( matches && ext === '.ogg' && isInAudioContainer ) {
+				return false;
+			}
+
+			return matches;
 		} );
 
 		const matchesKeyword = videoKeywords.some( ( keyword ) =>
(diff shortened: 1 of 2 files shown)


pattonwebz:
@coderabbitai Thanks for the nitpick on `src/pageScanner/checks/is-video-detected.js` (36-56) re: generalizing the audio-container exclusion beyond `.ogg` to cover `.webm` and other dual-purpose containers.

We are intentionally scoping this fix to `.ogg` only. `.ogg` is the extension actually reported in issue #1816 (the Gutenberg Audio block's default sample file), and Ogg Vorbis-only audio commonly keeps the `.ogg` extension in the wild. By contrast, audio-only WebM/MP4/ASF conventionally use dedicated extensions (`.weba`, `.m4a`, `.wma`) rather than the ambiguous ones already in `videoExtensions`, so the real-world false-positive risk there is much lower. We'd rather keep this PR narrowly focused on the reported bug than expand the ambiguous-extension list speculatively; we can revisit `.webm` (or others) in a follow-up if it turns out to be a real problem in practice.

PR #1821, opened 2026-07-05
Fix: keep sidebar actions menu button out of the header heading

### LINKED ISSUE #1763: Menu Button Is Incorrectly Exposed as a Heading
Location / URL: https://edac-vpat.instawp.co/wp-admin/post.php?post=1428&action=edit
Screenshot: 

Image

WCAG S.C: 1.3.1 Info and Relationships (Level A)

Issue Description: The Accessibility Checker actions menu button is announced by screen readers as "heading level 2" in addition to being identified as a collapsed menu button. The control is interactive and should not be exposed as a heading unless it is an actual heading in the page structure.

SteveJonesDev:
The Accessibility Checker actions menu button in the editor sidebar was rendered inside the `PluginSidebar` title, which Gutenberg wraps in an `<h2>`, so screen readers announced the button as "heading level 2" in addition to "collapsed menu button" (WCAG 1.3.1 Info and Relationships).

This PR passes a custom `header` to `PluginSidebar` instead: the `<h2>` now contains only the "Accessibility Checker" title text, with the busy spinner and the actions menu rendered as siblings outside the heading. The plain-text `title` prop is kept for the small-screen header, the Plugins menu entry, and as the rendered fallback if the `header` pass-through were ever removed. Title styles are now owned by the plugin (`.edac-sidebar__header-title`) instead of relying on core's internal `interface-complementary-area-header__title` class, which also let the two `:has()` overrides targeting core header classes be removed.

Fixes #1763

## Checklist

- [ ] Tests are added that cover changes — no existing JS test coverage for the sidebar registration component; verified via ESLint, webpack build, and the full Jest suite (842 tests passing). Worth a manual screen-reader check that the button no longer announces as "heading level 2".

https://claude.ai/code/session_013UoDxFBdso3isbuZXnm1hw

---
_Generated by [Claude Code](https://claude.ai/code/session_013UoDxFBdso3isbuZXnm1hw)_

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved sidebar header rendering for a cleaner, more consistent layout.
  * Ensured the sidebar title displays correctly with the loading indicator and action menu.
  * Updated title styling so long text is neatly truncated instead of wrapping awkwardly.

### DIFF
--- a/src/sidebar/index.js
+++ b/src/sidebar/index.js
@@ -90,15 +90,23 @@ function AccessibilityCheckerSidebar() {
 	return (
 		<PluginSidebar
 			name="accessibility-checker-sidebar"
-			title={
-				<span className="edac-sidebar__title">
-					{ __( 'Accessibility Checker', 'accessibility-checker' ) }
+			title={ __( 'Accessibility Checker', 'accessibility-checker' ) }
+			header={
+				// A custom header keeps the actions menu and spinner outside
+				// the header's <h2>, so screen readers don't announce the menu
+				// button as a heading. The plain-text `title` is still used by
+				// the small-screen header and the Plugins menu, and is the
+				// rendered fallback if `header` is ever unsupported.
+				<div className="edac-sidebar__header">
+					<h2 className="edac-sidebar__header-title">
+						{ __( 'Accessibility Checker', 'accessibility-checker' ) }
+					</h2>
 					{ backgroundRefresh && <Spinner className="edac-sidebar__title-spinner" /> }
 					<SidebarTitleMenu
 						postId={ postId }
 						refetchData={ refetchData }
 					/>
-				</span>
+				</div>
 			}
 			icon={ <AccessibilityCheckerIcon style={ { width: '24px', height: '24px' } } /> }
 		>
--- a/src/sidebar/sass/sidebar.scss
+++ b/src/sidebar/sass/sidebar.scss
@@ -40,22 +40,25 @@
 	}
 }
 
-.components-panel__header-title:has(.edac-sidebar__title) {
-	display: flex;
-	width: 100%;
-}
-
-.interface-complementary-area-header__title:has(.edac-sidebar__title) {
-	width: 100%;
-}
-
-.edac-sidebar__title {
+.edac-sidebar__header {
 	display: flex;
 	align-items: center;
 	gap: 8px;
 	width: 100%;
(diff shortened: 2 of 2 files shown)


src/sidebar/index.js
```diff
@@ -90,15 +90,23 @@ function AccessibilityCheckerSidebar() {
...
+				// the header's <h2>, so screen readers don't announce the menu
+				// button as a heading. The plain-text `title` is still used by
+				// the small-screen header and the Plugins menu, and is the
+				// rendered fallback if `header` is ever unsupported.
+				<div className="edac-sidebar__header">
+					<h2 className="edac-sidebar__header-title">
+						{ __( 'Accessibility Checker', 'accessibility-checker' ) }
+					</h2>
 					{ backgroundRefresh && <Spinner className="edac-sidebar__title-spinner" /> }
 					<SidebarTitleMenu
```
pattonwebz, line 105: **Regression: the "Pin to Toolbar" button is dropped**

`PluginSidebar`'s `header` prop, when supplied, entirely replaces the default header content rendered by core's `ComplementaryArea` — including the built-in "Pin to Toolbar"/"Unpin from toolbar" star button (see `packages/editor/.../components/complementary-area/index.js`: `children: header || <Fragment>{titleH2}{pinButton}</Fragment>`). Since `isPinnable` defaults to `true` and isn't overridden here, that button is still expected, but this custom `header` only renders the title, spinner, and `SidebarTitleMenu` — no pin control anywhere.

I verified this live: on `develop` the sidebar header shows a filled star "Unpin from toolbar" button next to the actions menu. On this branch (built + loaded in the block editor) that button is gone entirely, and `SidebarTitleMenu` doesn't offer a pin/unpin menu item either. So this fixes the heading-nesting bug but silently removes existing pin-to-toolbar functionality for every user.

Suggest re-adding a pin/unpin control inside the custom header — either reuse the same `isPinned`/`pinItem`/`unpinItem` wiring core's default header uses, or add a "Pin to toolbar" item to `SidebarTitleMenu`.

(Nice catch on also fixing the `title` prop — that incidentally fixes the toolbar toggle button's accessible name, which was rendering as `"[object Object]"` on `develop` since a JSX element was being passed where a string is expected.)

---

**Also: the title and action buttons are vertically misaligned**

Screenshot from 2026-07-10 18-36-02

PR #1846, opened 2026-07-25
Run make-pot on manual dispatch only, and drop the WordPress+MySQL scaffolding

pattonwebz:
## Why

`wp i18n make-pot` is pure static analysis over the source tree: it parses PHP/JS for translation calls and writes a `.pot`. It never bootstraps WordPress and never touches a database.

Verified rather than assumed — running it against this plugin on a machine with no WordPress, no MySQL and no `wp-config.php` exits 0. So this workflow's `mysql:5.7` service, readiness loop, wordpress.org tarball download, `wp-config` sed, `wp core install`, and the copy into `wp-content/plugins` and back out were all satisfying a dependency that does not exist.

## What changed

Net **−54 / +16 lines**, six steps removed:

- Gone: the `mysql:5.7` service, "Wait for MySQL to be ready", "Download and configure WordPress", "Install WordPress", the apt PHP install, and the `wp-cli.phar` download.
- `wp-cli` now comes from `shivammathur/setup-php`'s `tools:`, matching how `build-plugin-with-ref.yml` already sets up PHP in this repo.
- The `xdebug.max_nesting_level` line went with it: that apt line never installed xdebug, and `coverage: none` keeps it off.
- `--allow-root` dropped — the runner user isn't root.

## What deliberately did *not* change

- Same triggers, same `if:` gating on `automatic_translations_*` / `tm_edits_*`.
- Same `base` input and same PR targeting expressions.
- Same single `.pot` in `add-paths`. Nothing generates `.json` — `make-pot` writes one file, and JSON comes from the external translation system.
- `make-pot` still runs against `dist/accessibility-checker` and still writes into that copy's `languages/` folder, so **the generated POT is unchanged**.

## One detail worth reviewing

The committed POT at `dist/accessibility-checker/languages/accessibility-checker.pot` is *both* the file `make-pot` overwrites and the baseline the diff step compares against. So it's stashed to `$RUNNER_TEMP` before generation:

```bash
cd dist/accessibility-checker
if [ -f languages/accessibility-checker.pot ]; then
  cp languages/accessibility-checker.pot "${RUNNER_TEMP}/accessibility-checker.pot.baseline"
fi
wp i18n make-pot . languages/accessibility-checker.pot
```

Without that, the new POT would be compared against itself and `pot_changed` could never report `true`.

The source argument stays `.` (the plugin root) rather than the `languages/` folder, because `make-pot` resolves its `#:` source references against the *source argument*, not the working directory — verified by generating the same plugin from both locations and diffing: byte-identical.

## Testing

- `actionlint` clean; YAML parses.
- Verified zero remaining references to mysql, `wordpress/`, `latest.tar.gz`, `wp-config`, `wp core install`, `--allow-root`, `wp-cli.phar`, xdebug or `apt-get`.
- Replayed the generate + diff steps verbatim against the sibling `accessibility-new-window-warnings` plugin (small enough to run locally): the baseline is stashed byte-intact, the POT regenerates in place, `pot_changed=false` is correctly reported, and no `.po`/`.mo`/`.json` file is touched.
- The `make-pot` job reports `skipping` on this PR, since the branch isn't a translations branch. Worth dispatching once after merge to confirm it reports `pot_changed=false`.

Sibling PRs for `accessibility-checker-pro` and `accessibility-new-window-warnings` to follow. `accessibility-checker-multisite` has the same scaffolding *plus* a `wp i18n make-json` step, so it's left for a separate change.

https://claude.ai/code/session_019LoLgW7oFjBPxGea9kiZJ2

## Summary by CodeRabbit

* **Chores**
  * Updated the translation POT generation workflow to run as manual-dispatch only.
  * Simplified POT generation by removing the MySQL/WordPress sandbox and generating directly from the built distribution with PHP tooling.
  * Improved “POT changed” detection by diffing regenerated output against a stored baseline while filtering out selected header keys plus comment/blank lines.
  * Updated artifact handling to upload the newly generated POT from its new location and copy it to the commit-ready path.


pattonwebz:
Note for reviewers: CodeRabbit's two inline comments above are **stale** — they're anchored to commit `ec1bb803`, and the branch was force-pushed to `87dae184` after this PR was rescoped.

Both target code that no longer exists here. The original version of this PR also made the workflow accept arbitrary `ref`/`base` inputs; that feature was dropped (translations only ever run against `develop`, driven by dispatch and the externally-pushed `automatic_translations_*` / `tm_edits_*` branches), and it took the flagged code with it:

- **"Concurrency group doesn't mirror the `BASE` resolution precedence"** — there is no `concurrency:` block and no `inputs.ref` in the current file. It was a correct catch against the old version, where the concurrency key genuinely did skip `inputs.ref` and two dispatches resolving to the same base could have raced. Moot now.
- **"Swallowed `gh api` stderr"** — there is no `gh api` call anymore; the base-branch existence check went with the same feature.

Line 29 is now `- name: Checkout code` and line 112 is a `git config` line, so the anchors no longer point at related code.

The PR as it stands is a pure removal: `−54 / +16`, six steps deleted, no behaviour change to triggers, PR targeting, or the generated POT.

pattonwebz:
@coderabbitai full review

pattonwebz:
@coderabbitai full review

pattonwebz:
@coderabbitai full review

pattonwebz:
@coderabbitai full review

Issue #1863, opened 2026-08-05
Ambiguous link text (2.4.4) stops being flagged once a link's aria-label has extra text appended


pattonwebz:
A link with ambiguous link text (e.g. "read more", "click here") is correctly flagged as **Link Purpose (2.4.4) / ambiguous link text**. If the link then also gets an `aria-label` that includes some additional text appended to the original wording, the ambiguous-text flag disappears on the next scan — even though the link's purpose is no more clear than it was before.

This was reported by a user who hit it via the "Add Label To Links That Open A New Tab/Window" fix:

> there was an ambiguous link on the page that got flagged with "WCAG: 2.4.4 Link Purpose (In Context)," correctly, I believe. But when I applied "Add Label To Links That Open A New Tab/Window" to patch the [new-window] issue above, the 2.4.4 flag cleared on its own. It looks like the setting added an aria-label indicating the new tab, and that was enough to clear the check, even though the link purpose itself is still ambiguous.

The new-window fix is the reported trigger, but a theme, another plugin, or a hand-written `aria-label` may be able to produce the same result.

A link whose accessible name is still fundamentally ambiguous should continue to be flagged. Adding a new-window notice tells the user the link opens a new tab; it does not tell them where the link goes, so the original issue has not been resolved and the warning should not clear.

tw2113:
@pattonwebz tagging for assignment

amberhinds:
I tested and confirmed that there is a problem on https://a11ycheckrules.wpenginepowered.com/ambiguous-anchor-text/

Use the following code to test this. It should return **7 Link Opens New Window or Tab** review issues and **28 Ambiguous Anchor Text** problems. 

<details>
<summary>View Code</summary>

```
<!-- wp:heading -->
<h2 class="wp-block-heading">Passes</h2>
<!-- /wp:heading -->

<!-- wp:heading {"level":3} -->
<h3 class="wp-block-heading">String in Paragraph</h3>
<!-- /wp:heading -->

<!-- wp:paragraph -->
<p>This is a rule in the Accessibility Checker plugin by Equalize Digital. <a href="https://equalizedigital.com">To learn more about Equalize Digital click here.</a> The plugin is free on <a href="https://wordpress.org/plugins/accessibility-checker/">WordPress.org.</a></p>
<!-- /wp:paragraph -->

<!-- wp:heading {"level":3} -->
<h3 class="wp-block-heading">String in core button</h3>
<!-- /wp:heading -->

[...]

PR #1864, opened 2026-08-05
Retain ambiguous anchor text flags even if "Add Label To Links That Open A New Tab/Window" used

tw2113:
This PR handles the issue around [#1863 Ambiguous Text no longer being flagged despite not being fixed ](https://github.com/equalizedigital/accessibility-checker/issues/1863)

Changes include removal of truthy results for `aria-label` and `aria-labelledby` attribute checks, as well as removal of the Jest checks for those items. This enabled the checking of specifically the anchor link text or image alt text exclusively, for the passing state of Ambiguous text checks.

## Validation

Because we're now expecting to only check in the actual anchor text or image alt text, the tests for `aria-label` and `aria-labelledby` can be considered redundant.

Results from `npm run test:jest`:

Test Suites: 52 passed, 52 total
Tests:       944 passed, 944 total

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved ambiguous link-text detection by evaluating visible text and image alternative text.
  * Prevented accessible labels from being incorrectly flagged as ambiguous link text.

### DIFF
--- a/src/pageScanner/checks/has-ambiguous-text.js
+++ b/src/pageScanner/checks/has-ambiguous-text.js
@@ -34,12 +34,6 @@ const checkAmbiguousPhrase = ( text ) => {
 export default {
 	id: 'has_ambiguous_text',
 	evaluate: ( node ) => {
-		if ( node.hasAttribute( 'aria-labelledby' ) ) {
-			const label = node.getAttribute( 'aria-labelledby' );
-			const labelText = document.getElementById( label )?.textContent;
-			return checkAmbiguousPhrase( labelText );
-		}
-
 		if ( node.textContent && node.textContent !== '' ) {
 			return checkAmbiguousPhrase( node.textContent );
 		}
--- a/src/pageScanner/checks/has-ambiguous-text.js
+++ b/src/pageScanner/checks/has-ambiguous-text.js
@@ -34,11 +34,6 @@ const checkAmbiguousPhrase = ( text ) => {
 export default {
 	id: 'has_ambiguous_text',
 	evaluate: ( node ) => {
-		if ( node.hasAttribute( 'aria-label' ) ) {
-			const ariaLabel = node.getAttribute( 'aria-label' );
-			return checkAmbiguousPhrase( ariaLabel );
-		}
-
 		if ( node.hasAttribute( 'aria-labelledby' ) ) {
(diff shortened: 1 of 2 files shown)


tw2113:
@pattonwebz This is ready for your review

amberhinds:
Just noting here that this PR probably causes more issues than it fixes. I have concerns about the code deletions and think it needs extensive testing.

tw2113:
Definitely understand the concern there, but from what I could gather, removal is the only way to have the warning of the issue not being actually fixed yet still show up. So that aspect of some code removal is needed.

Willing to discuss as much as desired.

Would love to hear @pattonwebz's thoughts as well regarding the topic.

pattonwebz:
Testing the exact case here this does solve the specific problem. But it also removes the checks consideration of the aria-label as accessible name.

We probably need to solve this by making the label checking clever enough to understand that the longer label is still ambiguous even though it does not match one of our patterns exact. Might even be worth considering if the pattern matching should be loosed up to not be exact match anymore.

tw2113:
I suppose my question would be: should "ambiguous" still pass? should it pass but with a flagged warning of some sort?

I feel like there's a space between "longer ARIA label" succeeds and "Original Text" fails that could be explored. A "it sort of passes" state.

amberhinds:
I added example code that contains both pass and failure examples on the issue.

The reason you can't just remove checking the `aria-label` is that code like this needs to fail this rule.

`<a href="/hello-world" aria-label="read more">→</a>`

If you remove the aria-label check, then this would never fail.

We have to be careful not to flag it if the ambiguous text appears at the start of the aria-label, because that is correct.

`<a href="https://github.com/amberhinds" class="button" target="_blank" aria-label="Read more on Amber's GitHub" data-nww-processed="true">Read More</a>`

There may be a better way, but my initial thought is that we add "opens a new window" as a string that flags when combined exactly with our other strings (or on its own, in the case of an empty link).

amberhinds:
Follow up. This code does not fail for an empty link, but it does fail for ambiguous anchor text, so we have done something already to check for links that _only_ have a new window warning:
`<a href="https://github.com/equalizedigital" target="_blank" aria-label="opens a new window" data-nww-processed="true"><i class="edac-nww-external-link-icon" aria-hidden="true"></i></a>`

PR #1869, opened 2026-08-06
Fix: allow anchors with role none, presentation or slider

### LINKED ISSUE #1748: Rules: links with role none, role slider still flagging improper use of link
This code:
```
<a role="none"><span class="nav-drop-title-wrap">eBranch<span class="dropdown-nav-toggle"><span class="kadence-svg-iconset svg-baseline"><svg aria-hidden="true" class="kadence-svg-icon kadence-arrow-down-svg" fill="currentColor" version="1.1" xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24"><title>Expand</title><path d="M5.293 9.707l6 6c0.391 0.391 1.024 0.391 1.414 0l6-6c0.391-0.391 0.391-1.024 0-1.414s-1.024-0.391-1.414 0l-5.293 5.293-5.293-5.293c-0.391-0.391-1.024-0.391-1.414 0s-0.391 1.024 0 1.414z"></path> </svg></span></span></span></a>
```

Is still flagging Improper Use of link on Mountain Home (in the nav). Version 1.43.0 of Accessibility Checker.

Also on Highland (https://highland.edu/finearts/prairie-wind/), these links have a slider role:
```
<a class="mejs-horizontal-volume-slider edac-highlight-element-selected" href="javascript:void(0);" aria-label="Volume Slider" aria-valuemin="0" aria-valuemax="100" aria-valuenow="100" role="slider"><span class="mejs-offscreen">Use Up/Down Arrow keys to increase or decrease volume.</span><div class="mejs-horizontal-volume-total"><div class="mejs-horizontal-volume-current" style="left: 0px; width: 100%;"></div><div class="mejs-horizontal-volume-handle" style="left: 100%;"></div></div></a>
```

This is not an ideal implementation (an input would be better), but it is technically correct and common in older media players.

I would have thought that https://github.com/equalizedigital/accessibility-checker/pull/1698 might have fixed this, but maybe not.

mjoslyn:
Fixes #1748.

`link_improper` was flagging two anchors that are not exposed as links:

- `<a role="none">` / `<a role="presentation">` — removed from the accessibility tree entirely.
- `<a role="slider">` — exposed as a slider. Not ideal markup, as the issue notes, but common in older media players and technically correct.

`slider` joins the existing `button` / `tab` allowlist, which stays unconditional: none of those roles is exposed as a link, so the anchor is not being used improperly however it happens to be focused.

The presentational roles get a narrower exemption, following the [presentational roles conflict resolution](https://www.w3.org/TR/wai-aria-1.2/#conflict_resolution_presentation_none). The role is ignored — and the element therefore still exposed as a link — when the anchor is either:

- **Focusable.** Includes `tabindex="-1"`, which is reachable by click and programmatic focus even though it sits outside the tab order. So `<a href="#" role="none">` continues to fail.
- **Carrying a global ARIA state or property.** Approximated as any `aria-*` attribute, which errs toward reporting. `aria-hidden` is excluded, since it removes the element from the tree outright and makes the role moot — matching how `linked-image-alt-present` already handles it.

Also fixes an unrelated false positive found while working in this check: a `tabindex` that is not a valid integer is now ignored per the HTML spec rather than treated as focusable, so an empty named anchor like `<a id="top" tabindex="">` no longer reports.

### Compatibility

Every behavior change here is a relaxation. Nothing that passed on `develop` fails after this change, so it should not surface new issues on existing scans — only remove false positives.

Worth calling out one case I deliberately left alone: `<a role="button">` with no `href` and no `tabindex` still passes, as it does today. Reporting it is defensible, but it is a separate rule change from this issue and would need to account for `tabindex="-1"` being legitimate in composite widgets (an inactive tab in a roving-tabindex tablist, for instance). Happy to open a follow-up issue if that is wanted.

### Testing

15 cases added to `tests/jest/rules/linkImproper.test.js`, including both markup samples from the issue verbatim, the focusability and global-ARIA boundaries in both directions, and a roving-tabindex `role="tab"` case pinning the unconditional allowlist. Full file passes at 47/47; `phpcs` and `wp-scripts lint-js` are clean on the changed files.

## Checklist

### DIFF
--- a/includes/classes/Rules/Rule/LinkImproperRule.php
+++ b/includes/classes/Rules/Rule/LinkImproperRule.php
@@ -41,10 +41,12 @@ class LinkImproperRule implements RuleInterface {
 			),
 			'why_it_matters'        => esc_html__( 'Anchor tags (a.k.a. links) are intended for navigation to a new page or a different place on the same page. When they are used to trigger actions (such as expanding accordions or opening modals) without the correct roles or behavior, they confuse users, particularly those using screen readers or keyboards, who expect links to navigate rather than perform actions. They also are likely not to function with the space bar, which is an expectation of a button.', 'accessibility-checker' ),
 			'how_to_fix'            => sprintf(
-			// translators: %1$s is <code>&lt;button&gt;</code>, %2$s is <code>role="button"</code>.
-				esc_html__( 'If the element is used to trigger an action, replace the anchor tag with a %1$s. If you cannot replace it, ensure that %2$s is added to the link, along with JavaScript that adds support for triggering it with the space bar key, and that appropriate ARIA attributes are used for toggle states or other functionality.', 'accessibility-checker' ),
+			// translators: %1$s is <code>&lt;button&gt;</code>, %2$s is <code>role="button"</code>, %3$s is <code>role="none"</code>, %4$s is <code>role="presentation"</code>.
+				esc_html__( 'If the element is used to trigger an action, replace the anchor tag with a %1$s. If you cannot replace it, ensure that %2$s is added to the link, along with JavaScript that adds support for triggering it with the space bar key, and that appropriate ARIA attributes are used for toggle states or other functionality. If the anchor is purely decorative and cannot be focused, %3$s or %4$s marks it as presentational so it is not reported.', 'accessibility-checker' ),
 				'<code>&lt;button&gt;</code>',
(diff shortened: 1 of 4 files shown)


mjoslyn:
Closing to re-scope: reopening with a diff limited to the role exemptions in #1748.

PR #1871, opened 2026-08-07
Fix: label grouped settings fields with a legend

### LINKED ISSUE #1753: Checkbox Group Is Missing a <legend> Element
Location / URL: https://edac-vpat.instawp.co/wp-admin/admin.php?page=accessibility_checker_settings 

Screenshot: 

Image

WCAG S.C: 1.3.1 Info and Relationships (Level A)
Issue Description: The checkbox group is contained within a <fieldset>, but its group label ("Post Types To Be Checked") is implemented as a <label> associated with the first checkbox instead of a <legend>. As a result, the group label is not programmatically associated with the checkbox group, preventing assistive technologies from correctly conveying the relationship between the group label and its options.

mjoslyn:
Grouped checkbox and radio fields on the settings page passed `label_for` to
`add_settings_field`. WordPress renders that as a `<label>` bound to the *first*
control in the group, so the `<fieldset>` had no accessible name and the
group/options relationship was never exposed programmatically. Fails WCAG 1.3.1
Info and Relationships (Level A).

Fixes #1753

## Scope note

#1753 reports one group ("Post Types To Be Checked"). This fixes that group and
four others on the same page with the identical defect: Dismiss Permissions,
Prompt for Simplified Summary, Simplified Summary Position, and Frontend
Accessibility Checker Position. Same root cause and success criterion, so fixing
one and leaving four seemed worse than widening.

## What changed

- Drop `label_for` from the five grouped fields; titles now render as plain
  `<th>` text.
- Add a `screen-reader-text` `<legend>` to each `<fieldset>`, matching WP core.
  No visual change.
- Remove the ids those `label_for` args targeted, plus the dead first-item id
  special-cases they required. Ids are now keyed on post type / role slug rather
  than loop position.

The Settings API has no first-class way to name a group of controls — `label_for`
only ever targets a single input. Upstream:
https://core.trac.wordpress.org/ticket/39441

The other seven fieldsets in `options-page.php` each wrap a single control and
correctly keep `label_for`. Fieldsets elsewhere already have legends.

The radio groups now carry no id. Nothing consumed those ids —
`src/admin/index.js` selects by `name`. Two tests pin that down: group ids stay
unique, and every control keeps rendering the `name` it's addressed by.

## Testing

New `OptionsPageGroupLabelsTest`: 20 tests / 30 assertions. Full suite green
(897 tests, 1913 assertions, 15 pre-existing skips), `phpcs` clean.

Manually verified in `wp-env`: all five legends present and visually hidden, no
`<th>` still a label, zero duplicate ids page-wide, no layout shift.

## Not covered

- **Not verified with a real screen reader** — markup matches the core pattern,
  but announcement wasn't confirmed by hand.
- The name assertion pins the PHP side only; changing the selector in
  `index.js` would still break the toggle silently.

## Checklist

## Summary by CodeRabbit

* **Accessibility Improvements**
  * Improved screen-reader labeling for grouped checkbox and radio settings.
  * Added properly labelled fieldsets and legends to make grouped controls easier to understand and navigate.
* **Bug Fixes**
  * Ensured control IDs are unique and consistently associated with their corresponding post types and roles.
  * Removed redundant identifiers from select radio options.
* **Tests**
  * Added coverage for grouped settings markup, accessibility labels, control IDs, and input names.

### DIFF
--- a/includes/options-page.php
+++ b/includes/options-page.php
@@ -153,13 +153,14 @@ function edac_register_setting() {
 	);
 
 	// Add fields.
+
+	// No label_for: this field is a group of checkboxes labelled by the <legend> in its callback.
 	add_settings_field(
 		'edac_post_types',
 		__( 'Post Types To Be Checked', 'accessibility-checker' ),
 		'edac_post_types_cb',
 		'edac_settings',
-		'edac_general',
-		[ 'label_for' => 'edac_post_types' ]
+		'edac_general'
 	);
 
 	add_settings_field(
@@ -189,13 +190,13 @@ function edac_register_setting() {
 		[ 'label_for' => 'edacp_scan_all_taxonomy_terms' ]
 	);
 
+	// No label_for: this field is a group of checkboxes labelled by the <legend> in its callback.
 	add_settings_field(
 		'edacp_ignore_user_roles',
 		__( 'Dismiss Permissions', 'accessibility-checker' ),
 		'edac_ignore_user_roles_cb',
 		'edac_settings',
-		'edac_permissions',
-		[ 'label_for' => 'edac_ignore_user_roles' ]
+		'edac_permissions'
 	);
 
 	add_settings_field(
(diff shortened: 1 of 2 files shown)


mjoslyn:
@pattonwebz - ready for your review

BryanBelanger:
Resolved conflicts in preserving new permission tab and taxonomy fix alongside the four group legends. Also, updated to remove the obsolete Dismiss Permissions cases.

Validation:
- Tests and linting passed
- Browser smoke checks: labels, layout, radio keyboard interaction, and Permissions role eligibility passed
- Manual screen reader testing with VoiceOver to confirm legend announcements

PR #1899, opened 2026-08-24
Update sidebar store debounce and promise settlement

### LINKED ISSUE #1840: Sidebar refresh: module-level debounceTimer can permanently orphan an in-flight refetchData() promise
## Where this was spotted

Found during review of #1838 (\"Announce accessibility analysis refresh completion\"), which adds a `wp.a11y.speak()` announcement after the sidebar's Refresh action resolves.

## The issue

`debounceTimer` in `src/sidebar/store/accessibility-checker-store.js` is module-level state, shared across every call to the `refetchData` action — not scoped per-invocation:

```js
let debounceTimer = null;

refetchData( postId ) {
	return async ( { dispatch, select } ) => {
		if ( debounceTimer ) {
			clearTimeout( debounceTimer );
		}

		return new Promise( ( resolve ) => {
			debounceTimer = setTimeout( async () => {
				// ...resolve(...) only ever happens in here
			}, 200 );
		} );
	};
},
```

If `refetchData()` is called a second time within the 200ms debounce window (e.g. a fast double-click on the sidebar's "Refresh" menu item, or a background auto-refresh racing a user-triggered one), the second call's `clearTimeout(debounceTimer)` cancels the *first* call's pending timeout. Since `resolve()` only ever runs inside that now-cancelled callback, the first call's returned Promise never settles — it hangs forever.

## Why it was previously benign

Before #1838, nothing awaited or depended on this Promise settling — callers fired-and-forgot `refetchData()`, so an orphaned Promise for a superseded call was invisible and harmless. No behavior depended on it resolving.

## Why it's worth fixing now

#1838 makes `handleRefresh()` in `SidebarTitleMenu.js` `await` this Promise specifically to decide which `speak()` announcement to make (success vs. failure). If the Promise hangs due to the race above, `speak()` never fires for that click — the screen reader user gets **no announcement at all** for that interaction, rather than the guaranteed success/failure announcement the PR exists to provide. A user who double-clicks Refresh (plausible if unsure the first click registered) could end up with silent failure of the very feature meant to give them confirmation.

## Suggested fix direction

Either:
- Track the pending resolver(s) and resolve the superseded call (e.g. with `false`/a "superseded" signal) before clearing its timer, or
- Move away from a single shared `debounceTimer` toward tracking resolvers per in-flight call so none are silently dropped.

Not urgent/user-visible in the common case (requires a rapid double-trigger within 200ms), but worth fixing given the accessibility feature now depends on it.

BryanBelanger:
## Summary

Fixes: #1840 

The sidebar store debounces accessibility data refreshes by 200 milliseconds. Each refresh call also returns a promise so callers can wait for the refresh to complete.

When another refresh was requested during that debounce window, the existing timer was canceled and replaced. However, the promise associated with the canceled timer was never settled because its resolver was not retained. This could leave callers waiting indefinitely when rapid interactions or background refreshes triggered overlapping requests.

## Fix

This update stores the debounce timer and its corresponding promise resolver together as a pending refresh.

When a new refresh supersedes one that is still within the debounce window, the previous timer is canceled and its promise is settled with a `superseded` status. The latest refresh is then scheduled normally.

Once a refresh has started, it is no longer considered pending and is allowed to finish without being superseded by a subsequent request.

Jest coverage was added to verify that:

- A superseded refresh settles its promise without making an API request.
- Only the latest request within the debounce window is executed.
- A refresh that has already started is allowed to finish when another refresh is requested.

## Test plan

- Run `npm run lint:js`.
- Run `npm run test:jest`.
- Manually refresh the accessibility analysis from the available menu action and confirm the sidebar updates without hanging or becoming unresponsive.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved refresh handling so replaced refresh requests complete reliably instead of remaining pending.
  * Prevented unnecessary data fetches when a newer refresh supersedes an earlier one.
  * Preserved in-progress initial loading behavior during accessibility checks.

* **Tests**
  * Added coverage for refresh replacement, promise completion, and initial-load scenarios.


pattonwebz:
[Open Playground with this build](https://playground.wordpress.net/?blueprint-url=data%3Aapplication%2Fjson%3Bbase64%2CeyJsYW5kaW5nUGFnZSI6Ii93cC1hZG1pbi9wbHVnaW5zLnBocCIsInByZWZlcnJlZFZlcnNpb25zIjp7InBocCI6ImxhdGVzdCIsIndwIjoibGF0ZXN0In0sInN0ZXBzIjpbeyJzdGVwIjoibG9naW4ifSx7InN0ZXAiOiJpbnN0YWxsUGx1Z2luIiwicGx1Z2luWmlwRmlsZSI6eyJyZXNvdXJjZSI6InVybCIsInVybCI6Imh0dHBzOi8vbmlnaHRseS5saW5rL2VxdWFsaXplZGlnaXRhbC9hY2Nlc3NpYmlsaXR5LWNoZWNrZXIvYWN0aW9ucy9ydW5zLzMyNzU1NzU1MjAxL2FjY2Vzc2liaWxpdHktY2hlY2tlci0xLjQ4LjAtMTg5OS0xYzlhYzMyZC56aXAifSwib3B0aW9ucyI6eyJhY3RpdmF0ZSI6dHJ1ZX19XX0%3D)

Plugin path: accessibility-checker/accessibility-checker.php
Zip: https://nightly.link/equalizedigital/accessibility-checker/actions/runs/32755755201/accessibility-checker-1.48.0-1899-1c9ac32d.zip
Artifact: https://github.com/equalizedigital/accessibility-checker/actions/runs/32755755201/artifacts/9530801342
PR comment: not posted (missing permission - see the workflow log for details)

PR #1900, opened 2026-08-24
Add local context for issue dismiss in metabox area and resetting focus

### LINKED ISSUE #1771: Unexpected Focus Movement After Dismissing an Issue
Location / URL:https://edac-vpat.instawp.co/wp-admin/post.php?post=1639&action=edit&classic-editor__forget&classic-editor 
Screenshot: 

Image

WCAG S.C: 2.4.3 Focus Order (Level A)

Issue Description: When the Dismiss Issue button is activated, the corresponding issue is dismissed and the accordion collapses. After the content is updated, keyboard focus is automatically moved to the beginning of the results list instead of remaining at a logical location related to the user's previous interaction. This unexpected focus movement disrupts the navigation flow and requires keyboard and assistive technology users to navigate back to continue reviewing the remaining issues.

BryanBelanger:
## Summary

Fixes #1771 

When an issue was dismissed or reopened from the Accessibility Checker meta box, the details markup was rebuilt after the AJAX refresh. Because the focused element was removed from the DOM, keyboard users lost their focus position and context.

## Fix

This update captures the issue and rule-panel context before the action is submitted. After the details markup is refreshed it:

- Reopens the affected rule panel and issue controls.
- Restores the appropriate expanded state and ARIA attribute.
- Finds the replacement dismiss/reopen button and returns focus to it.

This keeps keyboard users in the same issue and panel after either action.

## Notes / follow-up considerations

- The button label changes slightly after the initial action.
- The current fix restores the panel associated with the most recent action. It does not preserve the complete expanded state of every open panel.
- If multiple panels are open and the user navigates elsewhere before opening another panel, restoring the acted-on panel can produce a layout shift. We may want to follow up by either preserving all expanded-panel state or collapsing unrelated panels during the refresh.

## Test plan

- Confirm linting and automated tests pass.
- Open an issue in the Accessibility Checker meta box using keyboard navigation.
- Place focus on the dismiss button and activate it.
- Confirm the affected rule and issue controls remain open after the refresh.
- Confirm focus returns to the corresponding reopen button.
- Repeat with multiple rule panels open and confirm the correct panel and button are restored.
- Verify the behavior with both keyboard navigation and mouse interaction.

## Checklist

## Summary by CodeRabbit

* **Accessibility Improvements**
  * Improved keyboard focus handling when issues are dismissed or reopened in the accessibility checker.
  * Preserved the relevant rule panel’s expanded state after updates.
  * Restored focus to the affected issue’s dismiss control after the panel refreshes.

### DIFF
--- a/src/admin/details/dismiss-issue-focus.js
+++ b/src/admin/details/dismiss-issue-focus.js
@@ -23,7 +23,7 @@ const findRuleDisplayBtn = ( root, panelId ) => {
 	}
 
 	const rulePanelDisplayButtons = Array.from( root.querySelectorAll( RULE_BUTTON_SELECTOR ) );
-	const rulePanelDisplayButton = rulePanelDisplayButtons .find( ( button ) => button.getAttribute( 'aria-controls' ) === panelId ) || null;
+	const rulePanelDisplayButton = rulePanelDisplayButtons.find( ( button ) => button.getAttribute( 'aria-controls' ) === panelId ) || null;
 
 	return rulePanelDisplayButton;
 };
@@ -92,7 +92,7 @@ export const restoreDismissIssueFocus = ( context, root = document ) => {
 		rulePanelIgnore.style.display = 'block';
 	}
 
-	const focusTarget = findIssueDismissBtn( root, context.issueId )
+	const focusTarget = findIssueDismissBtn( root, context.issueId );
 
 	focusTarget?.focus();
 
--- a/src/admin/details/dismiss-issue-focus.js
+++ b/src/admin/details/dismiss-issue-focus.js
@@ -1,0 +1,100 @@
+/**
+ * Focus helpers for a dismissed issue in the server-rendered details panel.
+ */
+
+const RULE_PANEL_SELECTOR = '.edac-details-rule-records';
+const RULE_BUTTON_SELECTOR = '.edac-details-rule-title-arrow';
+const RULE_TITLE_SELECTOR = '.edac-details-rule-title';
+const ISSUE_RECORD_SELECTOR = '.edac-details-rule-records-record';
+const ISSUE_RECORD_ID_PREFIX = 'edac-details-rule-records-record-';
+const ISSUE_RECORD_IGNORE_PREFIX = 'edac-details-rule-records-record-ignore-';
+const ISSUE_RECORD_DISMISS_BUTTON = '.edac-details-rule-records-record-ignore-submit';
+
+/**
+ * Find the rule button that controls a panel.
+ *
+ * @param {Document|Element} root    The root to search.
+ * @param {string}           panelId The controlled panel ID.
+ * @return {HTMLElement|null} The matching rule display button.
+ */
+const findRuleDisplayBtn = ( root, panelId ) => {
+	if ( ! panelId ) {
+		return null;
+	}
+
+	const rulePanelDisplayButtons = Array.from( root.querySelectorAll( RULE_BUTTON_SELECTOR ) );
+	const rulePanelDisplayButton = rulePanelDisplayButtons .find( ( button ) => button.getAttribute( 'aria-controls' ) === panelId ) || null;
+
+	return rulePanelDisplayButton;
+};
+
+/**
+ * Capture stable focus context for the issue being dismissed or reopened.
+ *
+ * @param {HTMLElement} submitButton The dismiss or reopen submit button.
+ * @return {Object|null} The dismissal focus context.
+ */
+export const captureDismissIssueFocusContext = ( submitButton ) => {
+	const issueRecord = submitButton?.closest( ISSUE_RECORD_SELECTOR );
+	const rulePanel = issueRecord?.closest( RULE_PANEL_SELECTOR );
+	const issueId = submitButton?.dataset?.id;
+
+	if ( ! issueRecord || ! rulePanel || ! issueId ) {
+		return null;
+	}
+
+	return {
+		issueId: String( issueId ),
+		rulePanelId: rulePanel.id,
+	};
+};
+
+/**
+ * Find an issue's dismiss button by its issue ID.
+ *
+ * @param {Document}    root    The document containing the metabox.
+ * @param {string|null} issueId The issue ID.
+ * @return {HTMLElement|null} The issue button.
+ */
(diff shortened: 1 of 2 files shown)


pattonwebz:
[Open Playground with this build](https://playground.wordpress.net/?blueprint-url=data%3Aapplication%2Fjson%3Bbase64%2CeyJsYW5kaW5nUGFnZSI6Ii93cC1hZG1pbi9wbHVnaW5zLnBocCIsInByZWZlcnJlZFZlcnNpb25zIjp7InBocCI6ImxhdGVzdCIsIndwIjoibGF0ZXN0In0sInN0ZXBzIjpbeyJzdGVwIjoibG9naW4ifSx7InN0ZXAiOiJpbnN0YWxsUGx1Z2luIiwicGx1Z2luWmlwRmlsZSI6eyJyZXNvdXJjZSI6InVybCIsInVybCI6Imh0dHBzOi8vbmlnaHRseS5saW5rL2VxdWFsaXplZGlnaXRhbC9hY2Nlc3NpYmlsaXR5LWNoZWNrZXIvYWN0aW9ucy9ydW5zLzMyNzU2MDgwMjU0L2FjY2Vzc2liaWxpdHktY2hlY2tlci0xLjQ4LjAtMTkwMC05ZjU4ODBmZi56aXAifSwib3B0aW9ucyI6eyJhY3RpdmF0ZSI6dHJ1ZX19XX0%3D)

Plugin path: accessibility-checker/accessibility-checker.php
Zip: https://nightly.link/equalizedigital/accessibility-checker/actions/runs/32756080254/accessibility-checker-1.48.0-1900-9f5880ff.zip
Artifact: https://github.com/equalizedigital/accessibility-checker/actions/runs/32756080254/artifacts/9530927449
PR comment: not posted (missing permission - see the workflow log for details)

src/admin/details/dismiss-issue-focus.js
```diff
@@ -0,0 +1,100 @@
...
+	if ( rulePanel ) {
+		rulePanel.style.display = 'block';
+
+		// Add active class to title and update aria on button
+		ruleDisplayBtn?.closest( RULE_TITLE_SELECTOR )?.classList.add( 'active' );
+		ruleDisplayBtn?.setAttribute( 'aria-expanded', 'true' );
+	}
+
+	if ( rulePanelIgnore ) {
+		rulePanelIgnore.style.display = 'block';
```
pattonwebz, line 92: This is opening the panel by setting the display to block but the aria-expanded is left as `false` in the markup. We should set that to true here, it should always be true in this path.

src/admin/index.js
```diff
@@ -232,6 +238,10 @@ const edacScriptVars = edac_script_vars;
 
 					// handle fix button click events.
 					initFixButtonEventHandlers();
+
+					if ( dismissFocusContext ) {
+						restoreDismissIssueFocus( dismissFocusContext );
```
pattonwebz, line 243: there is a return value from this call that sends back the element it focused or null. Can we use that value to keep people in the rule title arrow if the item is no longer found?

This could be a followup card if it's complicated to do here.

src/admin/details/dismiss-issue-focus.js
```diff
@@ -0,0 +1,100 @@
...
+	if ( ! context ) {
+		return null;
+	}
+
+	const rulePanel = root.getElementById( context.rulePanelId );
+	const ruleDisplayBtn = findRuleDisplayBtn( root, context.rulePanelId );
+	const rulePanelIgnore = root.getElementById( ISSUE_RECORD_IGNORE_PREFIX + context.issueId );
+
+	if ( rulePanel ) {
+		rulePanel.style.display = 'block';
```
pattonwebz, line 84: This same flipping happens elsewhere as well (I left a comment on it about toggling aria expanded on it). Could we make a helper to set it to block and to set the aria-expanded at once then call it in both places?

src/admin/details/dismiss-issue-focus.js
```diff
@@ -0,0 +1,100 @@
...
+		ruleDisplayBtn?.setAttribute( 'aria-expanded', 'true' );
+	}
+
+	if ( rulePanelIgnore ) {
+		rulePanelIgnore.style.display = 'block';
+	}
+
+	const focusTarget = findIssueDismissBtn( root, context.issueId );
+
+	focusTarget?.focus();
```
pattonwebz, line 97: question: does this focus also move the viewport for users that activate the dismiss with pointer?

src/admin/details/dismiss-issue-focus.js
```diff
@@ -0,0 +1,100 @@
...
+	const rulePanelDisplayButtons = Array.from( root.querySelectorAll( RULE_BUTTON_SELECTOR ) );
+	const rulePanelDisplayButton = rulePanelDisplayButtons.find( ( button ) => button.getAttribute( 'aria-controls' ) === panelId ) || null;
+
+	return rulePanelDisplayButton;
+};
+
+/**
+ * Capture stable focus context for the issue being dismissed or reopened.
+ *
+ * @param {HTMLElement} submitButton The dismiss or reopen submit button.
```
pattonwebz, line 34: The button might not actually be present if users don't have dismiss capabilities.

```suggestion
 * @param {HTMLElement|null} submitButton The dismiss or reopen submit button.
```

pattonwebz (changes requested):

src/admin/details/dismiss-issue-focus.js
BryanBelanger: Nope, not with my testing so far. Viewport has only been moving once I tab key into the "Dismiss issue as" radio options.

src/admin/details/dismiss-issue-focus.js
BryanBelanger: No problem, added an expandPanel helper to couple these updates together.

src/admin/index.js
BryanBelanger: Sure thing! Added another condition and leveraged findRuleDisplayBtn that helps handle this now.

src/admin/details/dismiss-issue-focus.js
BryanBelanger: This is now updated with the expandPanel helper.

PR #1905, opened 2026-08-27
Fix: textdomain loaded too early notice on plugins_loaded

pattonwebz:
## Summary
- A customer reported the "Translation loading for the accessibility-checker domain was triggered too early" notice from the Accessibility Checker (via a WordPress 6.7+ core check).
- `edac_ignore_capability()` was hooked to `plugins_loaded` and calls `edac_capability_metadata()`, which builds an array containing `__()` translation calls — this runs before `init`, before translations are loaded, triggering the core notice.
- Moved that hook, and the sibling `PermissionsPage::register_request_handlers()` registration (same file, same pattern), from `plugins_loaded` to `init`.
- Confirmed add-on plugins (Pro, Export, Audit History) register their `edac_capabilities` filter callbacks via `add_filter()` at plugin load time, so moving assembly to `init` doesn't change contribution order.

## Test plan
- [ ] `php -l includes/options-page.php`
- [ ] Confirm the "textdomain loaded too early" notice no longer appears with `WP_DEBUG`/`WP_DEBUG_LOG` enabled on a fresh page load
- [ ] Verify the Permissions settings tab still loads and saves correctly
- [ ] Verify add-on capabilities (Pro / Export / Audit History) still appear in the Permissions UI when those plugins are active

https://claude.ai/code/session_01Tz3KmgPopKSZYaPKTZFVRE

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved initialization timing for capabilities and permission handling.
  * Preserved the correct order of capability setup before permission processing.

### DIFF
--- a/includes/options-page.php
+++ b/includes/options-page.php
@@ -493,13 +493,17 @@ function edac_ignore_capability(): SyncCapability {
 	return $capability;
 }
 // Assemble after all plugins have loaded so add-ons can contribute to the
-// bundle via the filter regardless of plugin load order.
-add_action( 'plugins_loaded', 'edac_ignore_capability', 20 );
+// bundle via the filter regardless of plugin load order. Hooked to init
+// (rather than plugins_loaded) because edac_capability_metadata() calls
+// translation functions, and add-ons register their edac_capabilities filter
+// callbacks via add_filter() at plugin load time, so they're already in
+// place well before init fires either way.
+add_action( 'init', 'edac_ignore_capability', 20 );
 
 // Register the Permissions page request handler (admin-post save) on every
 // request. The tab UI itself is wired later on admin_menu.
 add_action(
-	'plugins_loaded',
+	'init',
 	function () {
 		$settings_capability = apply_filters( 'edac_filter_settings_capability', 'manage_options' );
 		( new PermissionsPage( $settings_capability ) )->register_request_handlers();
--- a/includes/options-page.php
+++ b/includes/options-page.php
@@ -441,7 +441,7 @@ function edac_sync_capability_roles(): void {
 }
 
 /**
- * The SyncCapability instance managing the bundle. Assembled on plugins_loaded
+ * The SyncCapability instance managing the bundle. Assembled on init
  * (see below) so every active add-on has contributed to edac_capability_bundle
  * first; also usable directly (e.g. in tests) as a lazy singleton.
  *
@@ -452,7 +452,7 @@ function edac_ignore_capability(): SyncCapability {
 
 	if ( null === $capability ) {
 		// Precompute each capability's floor once (all active add-ons have
-		// contributed metadata by the time this runs on plugins_loaded), so the
+		// contributed metadata by the time this runs on init), so the
 		// floor policy passed to the engine is a cheap array lookup rather than a
 		// metadata rebuild per role/capability during sync.
 		$floors = [];


pattonwebz (commented):
## Code Review Summary

**Verdict: Looks good** — 2 non-blocking suggestions

### What the fix does (correct)
Root cause is real: `edac_ignore_capability()` on `plugins_loaded` calls `edac_capability_metadata()`, which builds an array full of `__()` calls (lines 100–125) *before* translations load — triggering WP 6.7+'s "textdomain loaded too early" notice. Moving assembly to `init` (priority 20) is the right fix.

### Verified assumptions — all hold ✓
- **Ordering preserved:** add-ons contribute `edac_capabilities` via `add_filter()` at plugin load time (file-scope), so they're registered before `init` regardless of plugin load order. The lazy singleton (`static $capability`) means the bundle only assembles when actually needed; `init` is still early enough for admin-post saves (which fire after `init`).
- **No earlier callers:** `edac_ignore_capability()` is only invoked from `init` (line 501) and the sync helpers (364, 440) — no `plugins_loaded` path still assembles it.
- **Request cache:** `edac_capability_metadata()` has a static request cache, so floors compute once regardless of call sites.
- **Clean:** no stray `plugins_loaded` references remain; the comment block (495–500) accurately documents the *why*.

### 💡 Suggestions (non-blocking, see inline)
1. **Priority 21 is fragile** — the admin-post registration depends on the capability bundle assembling at priority 20. Consider one closure, or a comment noting the dependency.
2. **Consider a regression test** for the timing contract — nothing asserts assembly happens on `init`, not before.

### ✅ Looks good
- Minimal, focused diff
- Clear explanatory comments
- Lazy-singleton + cache pattern is clean
- All 14 CI checks green (lint PHP 7.4–8.2, integration WP 6.2/latest/multisite, security, style)

---
*Reviewed by Hermes Agent*

includes/options-page.php
```diff
@@ -493,13 +493,17 @@ function ( $role_slug, $cap ) use ( $floors ) {
...
 }
 // Assemble after all plugins have loaded so add-ons can contribute to the
-// bundle via the filter regardless of plugin load order.
-add_action( 'plugins_loaded', 'edac_ignore_capability', 20 );
+// bundle via the filter regardless of plugin load order. Hooked to init
+// (rather than plugins_loaded) because edac_capability_metadata() calls
+// translation functions, and add-ons register their edac_capabilities filter
+// callbacks via add_filter() at plugin load time, so they're already in
+// place well before init fires either way.
+add_action( 'init', 'edac_ignore_capability', 20 );
```
pattonwebz, line 501: 💡 **Suggestion (non-blocking):** Consider a regression test for the timing contract. `IgnoreCapabilityTest` covers the sync logic, but nothing asserts the bundle assembles on `init` (not `plugins_loaded`). A small test that fires `do_action('init')` and asserts `edac_ignore_capability()` was assembled — or asserts no `plugins_loaded` hook remains — would lock this fix in against regressions.

includes/options-page.php
```diff
@@ -493,13 +493,17 @@ function ( $role_slug, $cap ) use ( $floors ) {
...
 
 // Register the Permissions page request handler (admin-post save) on every
 // request. The tab UI itself is wired later on admin_menu.
 add_action(
-	'plugins_loaded',
+	'init',
```
pattonwebz, line 506: 💡 **Suggestion (non-blocking):** The priority `21` on this `init` hook (line 511) is fragile — this closure depends on `edac_ignore_capability()` assembling the capability bundle at priority `20` (line 501). If anything else ever hooks `init` at priority 20–21, the admin-post handler could register before the bundle exists. Consider combining both registrations into a single `init` closure, or adding a comment noting this dependency (e.g. "must run after priority 20, which assembles the capability bundle").

Issue #504, opened 2024-02-14
Test rule filters since JS scanning has been added


amberhinds:
Questions I don't know the answers to RE [How to Remove Rules (Advanced)](https://equalizedigital.com/accessibility-checker/how-to-remove-rules/):

1. If a rule is filtered out, does it delete the issues that already exist for it in the database?
2. When rules are filtered out, do all the reports in the free and pro hide that rule, or do they indicate it is disabled?
3. Can the Color Contrast rule be filtered out?
4. If the color contrast rule is filtered out does full site scanning say that it's only doing 1 pass? Or does the message to users still say pass 1 of 2.
5. What other things are notable about the experience when a rule is filtered out?

This isn't critical, but should be done before we rework wireframes for relevant sections, because we might need address issues/consistency around rule filtering and how filtered rules impact the admin.

amberhinds:
I have done some testing and here are the answers that I have so far after filtering out the ARIA-hidden rule:

**1. If a rule is filtered out, does it delete the issues that already exist for it in the database?**
No it's not deleting issues even if you re-scan the page and the issues in the database are still shown in the summary counts. Here's an example from my site:
<img width="1735" alt="Screenshot 2024-03-18 at 11 33 41 PM" src="https://github.com/equalizedigital/accessibility-checker/assets/13153456/f2888f23-f1a9-4f66-8236-d49e467244c4">

<img width="423" alt="Screenshot 2024-03-18 at 11 34 00 PM" src="https://github.com/equalizedigital/accessibility-checker/assets/13153456/c3e74e95-6577-463a-b5b4-9089c1959057">

There are only 5 warnings visible in the details tab but the summary still shows 15 because there are 10 aria-hidden warnings saved in the database.

**2. When rules are filtered out, do all the reports in the free and pro hide that rule, or do they indicate it is disabled?**

- The in-post report details tab hides the rule (not actually sure we should do this).
- The Open Issues: Individual Open Issues hides the rule.  (not actually sure we should do this)
- The code snippets still show in Fast Track though the check is missing:
<img width="1519" alt="Screenshot 2024-03-18 at 11 38 34 PM" src="https://github.com/equalizedigital/accessibility-checker/assets/13153456/3c1b67b5-171f-4fc8-a541-0eb171a16d49">
- Global Ignores still show up but with an empty check:
[...]

PR #1912, opened 2026-08-31
Fix: contrast error count ignores rule filtering

### LINKED ISSUE #1911: count_contrast_errors() ignores rule filtering, breaking error count math when color contrast is filtered
## Summary

`Summary_Generator::count_contrast_errors()` always queries the DB for `rule = 'color_contrast_failure'` and unconditionally subtracts that count from `errors` in `generate_summary()`, regardless of whether the color contrast rule has been filtered out via `edac_filter_register_rules`.

https://github.com/equalizedigital/accessibility-checker/blob/develop/includes/classes/class-summary-generator.php

```php
private function count_contrast_errors() {
	global $wpdb;

	$contrast_errors_count = $wpdb->get_var(
		$wpdb->prepare(
			'SELECT count(*) FROM %i where siteid = %d and postid = %d and rule = %s and ignre = %d',
			$wpdb->prefix . 'accessibility_checker',
			$this->site_id,
			$this->post_id,
			'color_contrast_failure',
			0
		)
	);

	return (int) $contrast_errors_count;
}
```

```php
$summary['contrast_errors']    = $this->count_contrast_errors();
$summary['errors']            -= $summary['contrast_errors'];
```

## Problem

Unlike the `errors`/`warnings` counts, this isn't just a "stale until rescan" issue — the query itself never checks whether `color_contrast_failure` is currently a registered/active rule. If color contrast is filtered out while violations still exist in the DB for a post, `generate_summary()` still subtracts the full contrast count from `errors`, which can produce an incorrect (including negative or understated) error total. This is independent of the rescan timing that resolves the other count-inflation issues.

## Reference

Found during investigation in equalizedigital/accessibility-checker#504 (see the code-level findings comment). That issue covered several other observations about rule filtering (no UI indication when a rule is disabled, general summary count inflation until rescan); this issue tracks specifically the `count_contrast_errors()` logic bug, which is the one that doesn't self-resolve.

## Suggested fix

Guard the subtraction (and/or the query) on whether `color_contrast_failure` is present in the currently active rule set from `edac_register_rules()`, similar to how filtered rules are excluded elsewhere.

SteveJonesDev:
## Summary

- `Summary_Generator::count_contrast_errors()` always queried the DB for `color_contrast_failure` violations and unconditionally subtracted the count from `errors` in `generate_summary()`, even when the color contrast rule has been filtered out via `edac_filter_register_rules`.
- Now takes the currently active `$rules` list and returns `0` (skipping the query and the subtraction) when `color_contrast_failure` isn't among them, so the error count stays correct while the rule is filtered and stale violations still exist in the DB.

## Why

Discovered during the rule-filtering investigation in equalizedigital/accessibility-checker#504. Unlike the general count-inflation behavior noted there (which self-resolves after a rescan), this was a logic bug in the query itself — filed and tracked separately as equalizedigital/accessibility-checker#1911.

## Test plan

- [x] Added two PHPUnit tests covering both branches (rule filtered → returns 0 without touching the DB result; rule active → counts violations as before)
- [x] Full suite run via `npm run test:php` / `test:php:run` — 968 tests pass
- [x] `./vendor/bin/phpcs` clean on changed files

Fixes equalizedigital/accessibility-checker#1911

## Summary by CodeRabbit

* **Bug Fixes**
  * Contrast violation totals are now calculated only when the color-contrast rule is enabled.
  * Reports correctly show zero contrast errors when that rule is excluded.

* **Tests**
  * Added coverage for enabled and disabled color-contrast rule scenarios.

### DIFF
--- a/includes/classes/class-summary-generator.php
+++ b/includes/classes/class-summary-generator.php
@@ -75,7 +75,7 @@ class Summary_Generator {
 		$summary['errors']             = $this->count_errors();
 		$summary['warnings']           = $this->count_warnings();
 		$summary['ignored']            = $this->count_ignored();
-		$summary['contrast_errors']    = $this->count_contrast_errors();
+		$summary['contrast_errors']    = $this->count_contrast_errors( $rules );
 		$summary['errors']            -= $summary['contrast_errors'];
 		$summary['content_grade']      = $this->calculate_content_grade();
 		$summary['readability']        = $this->get_readability( $summary );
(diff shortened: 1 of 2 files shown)


pattonwebz:
Non-blocking nits from review:

1. **Pattern inconsistency (minor):** `count_warnings()` / `count_ignored()` handle the filtered `link_blank` rule with a raw `rule != %s` clause inside the query, while this fix checks the rule list in PHP before querying. The PHP-side check is arguably cleaner and matches the issue's suggested approach — just flagging in case a future refactor wants to unify the two patterns.

2. **Cosmetic:** the `@param array $rules` docblock doesn't mention the `= []` default, which silently returns 0 when called with no args. Only reachable via reflection/tests today, so harmless.

Nice focused fix — the guard + both test branches cover it well.

PR #1929, opened 2026-09-12
fix(PRO-1332): stack issue title below the element-not-found notice

### LINKED ISSUE #1928: Front-End Highlighter: Move issue title below “element not found” notice
When an Accessibility Checker issue's element can no longer be found on the page, the “The element was not found on the page.” notice displays to the left of the issue title, if the issue title is show, such as with "Link to PDF".

Image

This causes the issue title (for example, “Link to PDF”) to appear on the same line as the notice and makes the layout less clear.

## Expected behavior

When the “element was not found” notice is present, the issue title should display on a new line below the notice, rather than beside it.

pattonwebz:
## Summary

When an issue's element can no longer be found on the page, the **"The element was not found on the page."** notice renders beside the issue title (e.g. "Link to PDF") instead of above it. The title should sit on its own line below the notice.

## Root cause

`.edac-highlight-panel-description-title` is a wrapping flex row (`display: flex; align-items: center; flex-wrap: wrap; gap: 8px`), and `descriptionOpen()` injects the notice as the **first flex item** in that row, immediately before the title `<span>` and severity badge. As a flex item it only wraps when the row runs out of space — so on a normal-width panel the notice and the title end up sharing a line, which reads as one run-on label.

## Fix

- New helper `src/frontendHighlighterApp/descriptionTitle.js` — `buildDescriptionTitle()` builds the notice/title/badge markup and reports whether a notice is present. Extracting it (mirroring the existing `getLandmarkType.js` + its Jest test) gives the layout rule a unit-testable seam.
- `index.js` uses the helper and flags the container with `edac-highlight-panel-description-title--has-notice` when a notice is present. The flag is removed in the "issues cleared" path alongside the existing `innerHTML` reset.
- `sass/app.scss` — the `--has-notice` modifier gives the notice a full flex line (`flex-basis: 100% !important`), so the title and badge wrap onto the next line while keeping their own alignment. Scoped to the notice, so the no-notice layout is untouched.

This covers all three notice states that share the markup (not found / not visible / not focusable), not just the "not found" case in the report.

## Test plan

- [x] Jest: `npm run test:jest tests/jest/frontendHighlighterApp/descriptionTitle.test.js` — new coverage asserts the notice is rendered before the title, `hasNotice` is true/false appropriately, the heading role is preserved, and the badge follows the title.
- [ ] Manual: open the front-end highlighter on a page where an issue's element is gone (e.g. a "Link to PDF" issue whose target was removed) → notice on its own line, issue title below it.
- [ ] Manual: an issue whose element *is* present renders unchanged (title + badge on one line, no notice).
- [ ] Manual: "rescan this page" after clearing issues → no stale `--has-notice` styling on the emptied panel.
- [x] `npm run lint:js` and `npm run lint:css`.

> **Note on local verification:** this environment has no `node_modules` (Docker wp-env not available), so Jest/PHPCS/ESLint could not be run locally. Verified here: `node --check` on all three JS files, the helper's behaviour mirrored 1:1 against the new Jest expectations (7/7), and SCSS brace balance (169/169). CI is the source of truth for the Jest + lint jobs.

- Linear: PRO-1332
- Fixes #1928

## Summary by CodeRabbit

* **New Features**
  * Issue descriptions now display status notices alongside titles when applicable.
  * Severity badges are included in issue title markup.

* **UI Improvements**
  * Status notices appear on their own line, keeping issue titles clearly separated and readable.
  * Clearing an issue now removes notice-specific title styling.
  * Titles maintain consistent heading structure and badge placement across different notice states.


amberhinds:
Tested and confirmed this works.

